PR #1, opened 2019-09-04
README: update to use out of tree build

merlinND:
I believe that out-of-tree build is the recommended way when using CMake, I propose we update the instructions to reflect that.

I also removed the core count argument for `make -j` because in my experience it always set it automatically to a sensible default, let me know if `4` is a better number.


wjakob:
Closing ancient bug..

PR #13, opened 2020-03-05
[bsdfs/roughconductor] Include specular_reflectance component

jammm:
Some scenes (such as glass-of-water) require a custom specular_reflectance component to be set.
As this is my first PR, please let me know if I made any mistake :-)

### DIFF
--- a/src/bsdfs/roughconductor.cpp
+++ b/src/bsdfs/roughconductor.cpp
@@ -231,7 +231,7 @@ public:
         // Ensure that this is a valid sample
         active &= neq(bs.pdf, 0.f) && Frame3f::cos_theta(bs.wo) > 0.f;
 
-        Float weight;
+        UnpolarizedSpectrum weight;
         if (likely(m_sample_visible))
             weight = distr.smith_g1(bs.wo, m);
         else
@@ -279,7 +279,7 @@ public:
 
         /* If requested, include the specular reflectance component */
         if (m_specular_reflectance)
-            F *= unpolarized<Spectrum>(m_specular_reflectance->eval(si, active));
+            weight *= m_specular_reflectance->eval(si, active);
 
         return { bs, (F * weight) & active };
     }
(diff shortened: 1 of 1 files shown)


wjakob:
Hi @jammm,

this looks good to me, but I had one small request: could you make the new ``specular_reflectance`` parameter `nullptr` by default (analogous to the changes here: https://github.com/mitsuba-renderer/mitsuba2/commit/a3531ac4245cb60017f45b4eb67a4522ee5c5d1e). It is a non-physical parameter (specular reflectance is determined by the Fresnel term, anything else is suspect and should  probably modeled in another way :-)), and I would prefer it to completely disappear from the system unless explicitly requested.

Thanks,
Wenzel

jammm:
Hi @wjakob,

Thanks a lot for the feedback!
I made the changes you requested. Can you please take a look?

Cheers,
Aaryaman

src/bsdfs/roughconductor.cpp
```diff
@@ -274,6 +277,10 @@ class RoughConductor final : public BSDF<Float, Spectrum> {
             F = fresnel_conductor(UnpolarizedSpectrum(dot(si.wi, m)), eta_c);
         }
 
+        /* If requested, include the specular reflectance component */
+        if (m_specular_reflectance)
+            F *= unpolarized<Spectrum>(m_specular_reflectance->eval(si, active));
```
wjakob: Change `weight` above to type `UnpolarizedSpectrum` & change this line to ` weight *= m_specular_reflectance->eval(si, active);
`.

src/bsdfs/roughconductor.cpp
```diff
@@ -346,6 +353,10 @@ class RoughConductor final : public BSDF<Float, Spectrum> {
             F = fresnel_conductor(UnpolarizedSpectrum(dot(si.wi, H)), eta_c);
         }
 
+        /* If requested, include the specular reflectance component */
+        if (m_specular_reflectance)
+            F *= unpolarized<Spectrum>(m_specular_reflectance->eval(si, active));
```
wjakob: Change `result` above to type `UnpolarizedSpectrum` & change this line to ` result *= m_specular_reflectance->eval(si, active);
`.

wjakob:
Two more small suggestions regarding polarization, the rest looks good.

jammm:
Thanks !

I made the requested changes. Should I squash them all into a single commit?

jammm:
Done! Please take a look :)

wjakob:
Thanks, I can also do it :).

PR #45, opened 2020-03-25
Add xml.load_dict() to load objects from py::dict

arpit15:
## Added functionality
- Access to plugin manager instance
- example showing how to create a scene in python

## Limitation
Only supports mitsuba tags which are python constructible(primitive types, shape, integrator, emitter, scene). 
The limiting factor is the requirement of `base class ptr` in the call to `create_object`.  Due to this limitation, there is a lot of unnecessary code during python scene creation.


Speierers:
Thanks for this PR @arpit15 !

Python bindings for the `PluginManager` would be a great feature to have in Mitsuba 2!

Here is a snippet of Python code showing what we had in mind regarding how the `PluginManager` would be used in Python:

```python
import mitsuba
mitsuba.set_variant('scalar_rgb')
from mitsuba.core import *

pmgr = PluginManager.instance()
camera = pmgr.create({
    "type" : "perspective",
    "to_world" : Transform.look_at(
        Point3f(0, 0, -10),
        Point3f(0, 0, 0),
        Vector3f(0, 1, 0)
    ),
    "film" : {
        "type" : "hdrfilm",
        "width" : 1920,
        "height" : 1080
    }
})
```

It can't be much cleaner than that, and it matches closely the structure of XML scene files. Let's try to aim for this in the long run :wink:

This will be quite a bit of work, involving changes in various places of the codebase. 

- [ ] `Properties` from `dict` Python constructor
- [ ] Recursive instantiation of objects from `dict`
- [ ] Implement [Spectra converstion](https://mitsuba2.readthedocs.io/en/latest/generated/plugins/#spectra)
- [ ] Bindings for the `PluginManager` class

Your help on this would be really appreciated and I am happy to contribute as well!

src/libcore/python/object.cpp
```diff
@@ -30,7 +30,8 @@ class PyTraversalCallback : public TraversalCallback {
 };
 
 MTS_PY_EXPORT(Object) {
-    py::class_<Class>(m, "Class", D(Class));
+    py::class_<Class>(m, "Class", D(Class))
+        .def_method(Class, parent, py::return_value_policy::reference);
```
Speierers: Let's not expose this low-level logic to Python. The `PluginManager.create(..)` should be able to handle this.

src/libcore/python/plugin_v.cpp
```diff
@@ -0,0 +1,13 @@
...
+#include <mitsuba/python/python.h>
+
+MTS_PY_EXPORT(PluginManager) {
+  MTS_PY_IMPORT_TYPES()
+  py::class_<PluginManager, std::unique_ptr<PluginManager, py::nodelete>>(m, "PluginManager")
+    .def_static_method(PluginManager, instance, py::return_value_policy::reference)
+    .def_method(PluginManager, ensure_plugin_loaded, "name"_a)
+    .def_method(PluginManager, loaded_plugins)
+    .def_method(PluginManager, register_python_plugin, "plugin_name"_a)
+    .def("create_object", (ref<Object> (PluginManager::*)(const Properties &props, const Class *class_)) &PluginManager::create_object, "props"_a, "class_"_a, py::return_value_policy::reference);
```
Speierers: I guess the bindings here shouldn't be a one-to-one matching API with the cpp API.

I would just have bindings for `instance()` and `create(..)` for now.
Also, `create(..)` will have to be implemented here (using a `lambda function`) and it will call the `create_object` method once that `Properties` has been created from the dictionary. Something like this:

```cpp
.def("create", [](py::dict &args){
    Properties props = props_from_dict(args);
    return create_object(props, ...);
}, "args"_a)

Comment (arpit15) [reply src/libcore/python/plugin_v.cpp]: This is the syntax that I want to achieve. However, I can't figure out how to create a ptr to the base class to make a call to the plugin manager with the given props. 
For ex: to create a sensor using the dict in your example for lambda function, how would you obtain `Sensor` class to make a call to `PluginManager::create_object`?

Comment (Speierers) [reply src/libcore/python/plugin_v.cpp]: I would do it the same way it is done in the `xml.cpp`:
```
PluginManager::instance()->create_object(
        props,
        Class::for_name("Sensor", mitsuba::detail::get_variant<Float, Spectrum>()));
```
Is this what you were looking for?

Comment (arpit15) [reply src/libcore/python/plugin_v.cpp]: I tried `for_name` with `integrator` only. Thanks for pointing out that I need to capitalize for the class. I will try to use the above suggested fix and submit updates.

File: src/libcore/python/plugin_v.cpp [hunk 3]
```diff
@@ -1,13 +1,98 @@
 #include <mitsuba/core/plugin.h>
 #include <mitsuba/core/properties.h>
+#include <mitsuba/core/spectrum.h>
+#include <mitsuba/core/transform.h>
 #include <mitsuba/python/python.h>
 
+void parse_spectrum(PluginManager &pgmr, Properties &prop, py::dict &dict) {
```
Speierers: As a general comment, it would be great to use 4 spaces for indentation so to be consistent with the rest of the codebase 😉

Speierers:
Looking at your changes, it seems like having all of those `key.compare(..)` is not the right approach as all plugins have different properties with different names. I would therefore suggest taking the following assumptions (not sure those are perfect, but we can iterate from here then):

**1.** Sub-dictionaries are only for `mitsuba::Object`, the key must be the interface name and they must contains a `type` entry:

_Skeleton_:
```python
"interface_a" : {
    "type": "type_a",
    "interface_b": {
        "type": "type_b",
         ...
    },
    ...
}
```
_Example_:
```python
"shape" : {
    "type": "obj",
    "emitter": {
        "type": "area",
         ...
    },
    ...
}
```


**2.** Other properties (e.g. `to_world`, `sample_count`, ...) will have their type deduced from the variable given and the key will be there name:

_Skeleton_:
```python
"interface" : {
    "type": "type",
    "property_name": python_object_property_variable,
    ...
}
```

_Example_:
```python
some_transform = Transfrom3f(..)

"shape" : {
    "type": "obj",
    "to_world": some_transform,
    ...
}
```

So as you iterate over the dictionary, it won't be necessary to do all of those `key.compare()` as you will know how to handle a (key, value) pair based on the two assumptions above. This should make the code much more elegant and will work with any type/name of properties.

Let me know if you agree with the design! Happy to discuss it further with you.

arpit15:
Thanks for reviewing it. It would be great to incorporate the changes based on the assumptions you suggested. I have following doubts/help required regarding them:
1. Can you suggest a code snippet to set `Properties::Type` without explicitly checking types?
2. Even if I assume `type` is a key in the dict, I would have to iterate over all the elements and check if the current key is not `type`. I can get away with comparing for `type` only if I assume that it is always the first key in the passed dict. 
3. `key.compare()` checks for `spectrum` and `rgb` will remain as they are handled separately.  Can you suggest a way to get rid of them.

Speierers:
> 1. Can you suggest a code snippet to set Properties::Type without explicitly checking types?

You could use the `typeid(T).name()` as follow:

```c++
    // Helper macro
    #define SET_PROP(Type, Setter)                                        \
        if (strcmp(typeid(T).name(), typeid(Type).name()) == 0)           \
            props.Setter(name, value);

    // NOTE: we might want to add this function directly to Properties actually
    template <typename T>
    void set(Properties &props, const std::string &name, const T &value) {
        SET_PROP(bool, set_bool)
        SET_PROP(int, set_int)
        SET_PROP(long, set_long)
        SET_PROP(float, set_float)
        SET_PROP(std::string, set_string)
        SET_PROP(Vector3f, set_vector3f)
        SET_PROP(Point3f, set_point3f)
        SET_PROP(Transform4f, set_transform)

        // TODO: not sure about those two
        SET_PROP(ref<AnimatedTransform>, set_animated_transform)
        SET_PROP(ref<Object>, set_object)
    }

#undef SET_PROP
```
I haven't tried to compile this but you can find similar code in `object_v.cpp`. 

`Object` might need to be treated differently here, but you could already try the simpler case, e.g. `Vector3f` of `float`.

> 2. Even if I assume type is a key in the dict, I would have to iterate over all the elements and check if the current key is not type. I can get away with comparing for type only if I assume that it is always the first key in the passed dict.

For now let's assume `type` is always the first key in the subdictionary. Otherwise throw an error with an explicit error so the user know how to fix his mistake. We might be able to relax that assumption later, let's see.

> 3. key.compare() checks for spectrum and rgb will remain as they are handled separately. Can you suggest a way to get rid of them.

Indeed `spectrum` and `rgb` should be handled differently. I would focus on the rest for now, this will require a bit more thinking. It should probably share some logic with the XML parser to avoid code duplication.

arpit15:
@Speierers I am still struggling in implementing your suggestion regarding types. Currently, I have following snippet which is not able to currently parse the types. The type inferred for `T` is `py::handle` which leads to it not matching any `Properties::Type`. `object_v.cpp` seems to take type information as arguments. 

Could you guide me with some more hints?
```cpp
#define SET_PROP(Type, Setter)                                      \
  if (strcmp(typeid(T).name(), typeid(Type).name()) == 0) {       \
    Type ptr = py::cast<Type>(value);               \
    props.Setter(name, ptr);                  \
    return;                           \
  }

template <typename T>
void set(Properties &props, const std::string &name, const T &value) {
  SET_PROP(std::string, set_string);
}

void myfunc(py::dict dict) {
  Properties props;
  for(auto it : dict) {
    auto key = it.first.cast<std::string>();
    set(props, key, it.second);
  }
}
```

```python 
mydict = {
  "shape" : {
    "type" : "rectangle"
  }
}

myfunc(mydict)
```

Speierers:
Okay, in this case it is different as we are working with `py::dict` which gives us `py::handle`.

Let's try something like this instead:

```c++
#define SET_PROP(Type, Setter)                                        \
    try {                                                             \
        Type tmp = it.second.cast<Type>();                            \
        props.Setter(key, tmp);                                       \
    } catch (const pybind11::cast_error &e) { }
```

```c++
m.def("fooo", [](py::dict dict){
    Properties props;
    for(auto it : dict) {
        auto key = it.first.cast<std::string>();
        SET_PROP(std::string, set_string)
        SET_PROP(float, set_float)
        // ...
    }
    std::cout << "props: " << props << std::endl;
});
```

This works for me with:
```python
mydict = {
   "type" : "rectangle",
   "value" : 1.0
}
mitsuba.core.fooo(mydict)
```

arpit15:
Thanks for the quick reply. There are 2 problems with making a `Properties` cstr using `py::dict`
1. It will make codebase dependent on `pybind11`, even if python bindings are not requested
2. `Properties` is not recursive. Therefore, `properties.cpp` will have to contain parsing code which is currently in `xml.cpp`. It will also have to depend on `plugin_manager.cpp` to create obj.

Speierers:
I meant a Python constructor, which would only exist in the bindings, so I don't fully agree with `1.`.

But you have a point, let's stick with the original design for now!

Speierers:
Actually, a more efficient way of casting a `py::handle` to the right type would be to do the following:
```c++

#define SET_PROP(PyType, Type, Setter)                                   \
    case &PyType:                                                        \
        props.Setter(key, it.second.cast<Type>()); 

...

PyTypeObject *type = (PyTypeObject *) it.second.get_type();
switch (type) {
   SET_PROP(PyFloatObject, float, set_float);
   SET_PROP(PyLongObject, int, set_int);
   ...
}
```

You can find more about this here:
https://docs.python.org/3/c-api/concrete.html

Although I suggest we first try to get it working with the `try .. catch` macro. We can always come back to this later.

arpit15:
I tried `try .. catch`. However, without explicitly type checking for python types, the values are casted into a different types. For ex: `float` can be converted to `bool` and vice-versa. This leads to problems in declaring the following 

```
"shape" : {
    "type" : "sphere",
    "radius" : 10.0,
    "flip_normals" : False
}
```
Currently the following works fine 
```
#define SET_PROP(Type, Setter)                  \
if(isinstance<Type>(val)) {                 \
  Type tmp = val.cast<Type>();                            \
      prop.Setter(key, tmp);                                  \
      continue;                       \
  }
```
Therefore, it seems that I will have to either use what you suggested or check the instance like in my current code. Let me know which one do you like. I can easily change it. 
I have a question regarding spectra. Where can I find all the variations allowed for spectra. [This](https://mitsuba2.readthedocs.io/en/latest/generated/plugins.html#spectra) section lists variation in which there is an only a single XML statement. However, when I look at `xml.cpp` it seems that in some cases `RGB` can be a child of `Spectrum` as its parsing code takes `within_spectrum` bool. Can you give some snippets or links to models which use such a syntax.

Speierers:
You code snippet with the `isinstance` looks good to me. Let's stick with that for now.

I will have a look at `xml.cpp`, not sure to be aware of this special case for spectrums. I will get back to you ASAP

Speierers:
@arpit15 any progress on your side?

I would like to take a stab at this as it might be helpful in a future refactoring of the test suite. Would you mind giving me push-rights to your branch so I can contribute to this task?

Thanks

arpit15:
@Speierers I updated the branch and gave you push rights on my repo. During implementation, I noticed the following problems which I haven't had the time to resolve
1. In the current format, there is no way of creating 2 objects of the same type using 1 dict. This is creating a problem specifically in creating a `checkerboard` texture. 
2. Texture had to be handled separately. The problem is common for any named object. A simple solution might be to return the name from the `create_properties` func if the dict contains one and use it as a key to add the object into the properties instead of `parent_class`

Speierers:
Thank you @arpit15 !

I just pushed some of the changes I made today:

- Moved everything to `xml::load_dict` as it isn't really related to the `PluginManager` itself.
- Load regular and irregular spectrum
- and more...

Right now there is a lot of duplicated code from `xml.cpp`, especially for parsing rgb and spectrum. I am trying to get rid of this.

Speierers:
One last thing to be done: handle `point3f` and `vector3f` differently in `Properties` so one can set a `Point3f` and later query a `Vector3f`.

Speierers:
Hi @arpit15, 
I just rebase the latest commits from master and squash the commits. For this I had to force push so be careful next time you pull from this branch 😉

Speierers:
Ready for reviews!

Speierers:
Behaviour changed to support multiple objects with the same interface.

src/libcore/tests/test_xml.py
```diff
@@ -297,3 +297,269 @@ def test20_upgrade_tree(variant_scalar_rgb):
...
+        }
+    })
+
+    e2 = xml.load_string("""
+        <emitter type="point" version="2.0.0">
+            <spectrum name="intensity" value="0.44"/>
+        </emitter>
+    """)
+
+    assert str(e1) == str(e2)
```
merlinND, line 480: I'm not sure if that's done indirectly here, but it would be nice to check explicitly that all the processing (pre-integration or not, within emitter, etc) is done correctly (both RGB & spectral mode, maybe mono as well).

include/mitsuba/core/properties.h
```diff
@@ -54,6 +53,7 @@ class MTS_EXPORT_CORE Properties {
     };
 
     using Float = float;
```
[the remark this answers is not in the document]
Speierers: I wouldn't change this. It would make the code a bit heavy, and this is also done in other files (e.g. `bitmap.h`, `transform.h`, `spiral.h`, `xml.cpp`, ...)

src/libcore/python/xml_v.cpp
```diff
@@ -38,8 +47,160 @@ MTS_PY_EXPORT(xml) {
...
+    Throw("Missing key 'type' in dictionary: %s", dict);
+}
+
+#define SET_PROPS(PyType, Type, Setter)         \
+    if (py::isinstance<PyType>(value)) {        \
+        props.Setter(key, value.cast<Type>());  \
+        continue;                               \
+    }
+
+MTS_VARIANT ref<Object> load_dict(const py::dict &dict) {
```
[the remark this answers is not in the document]
Speierers: Any specific use cases in mind? 

As you know, C++ isn't as easy with types compare to Python. Not sure it would be that easy to make it compatible with `std::map`.

include/mitsuba/core/properties.h
```diff
@@ -206,19 +206,12 @@ class MTS_EXPORT_CORE Properties {
     /// Retrieve a named reference value (use default value if no entry exists)
     const NamedReference& named_reference(const std::string &name, const NamedReference &def_val) const;
 
-    /// Store a 3D vector in the Properties instance
```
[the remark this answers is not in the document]
Speierers: `pros.vector3f()` was only used in a single place in the entire codebase, not really sure it is worth keeping distinct accessors. I would find it more confusing to now have `vector3f`, `point3f` and `array3f` as setters and accessors on `Properties`.

src/libcore/python/xml_v.cpp
```diff
@@ -38,8 +47,160 @@ MTS_PY_EXPORT(xml) {
...
+
+    std::string type = get_type(dict);
+    Properties props(type);
+
+    // TODO unfortunately this has to be done by hand now
+    bool within_emitter = (type == "area") || (type == "constant") ||
+                          (type == "directional") || (type == "envmap") ||
+                          (type == "point");
```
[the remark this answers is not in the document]
Speierers: Indeed! I will try to extend the `PluginManager` API to get ride of this

include/mitsuba/core/xml.h
```diff
@@ -40,5 +40,27 @@ extern MTS_EXPORT_CORE ref<Object> load_string(const std::string &string,
...
+
+/// Create a Texture object from a constant value or spectral values if available
+extern MTS_EXPORT_CORE ref<Object> create_texture_from_spectrum(
+                                        const std::string &name,
+                                        float const_value,
+                                        std::vector<float> &wavelengths,
+                                        std::vector<float> &values,
+                                        const std::string &variant,
+                                        bool within_emitter,
+                                        bool is_spectral_mode,
```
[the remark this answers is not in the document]
Speierers: We can't really template this as `xml.cpp` isn't templated. This function being in the `detail` namespace, I think it is fine for it to have "heavy" arguments like this.

I would prefer not having to expose the `ColorMode` enum in `xml.h` (to share it with `xml_v.cpp`).

src/libcore/python/xml_v.cpp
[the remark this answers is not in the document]
Speierers: It could be useful to add a `create_object()` method to `Properties` directly, so we could build a `Properties` in python and later construct the corresponding object. What do you think?
This would mean exposing the `PluginManager::create_object()` function. Although we could have it live directly on `Properties` for simplicity.

src/libcore/python/xml_v.cpp
```diff
@@ -38,8 +47,160 @@ MTS_PY_EXPORT(xml) {
...
+        // Cast py::list and numpy.ndarray to Array3f
+        if (py::isinstance<py::list>(value) ||
+            strcmp(value.ptr()->ob_type->tp_name, "numpy.ndarray") == 0) {
+            if (PyObject_Length(value.ptr()) != 3)
+                Throw("Expect array of size 3, found: %u", PyObject_Length(value.ptr()));
+            props.set_array3f(key, value.cast<Properties::Array3f>());
+            continue;
+        }
+
+        // Load nested dictionary
```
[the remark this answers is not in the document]
Speierers: I will add a comment for this

src/libcore/tests/test_xml.py
```diff
@@ -297,3 +297,269 @@ def test20_upgrade_tree(variant_scalar_rgb):
...
+                "height" : 768,
+            },
+            "mysampler" : {
+                "type" : "independent",
+                "sample_count" : 4,
+            },
+        },
+        "myemitter" : {"type" : "constant"},
+        "myshape" : {"type" : "sphere"}
+    })
```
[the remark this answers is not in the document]
Speierers: > * A random unreferenced `<sampler>` or `<film>` at the top level.

What do you mean by this? This is valid in XML at least.

Speierers:
Thanks @merlinND for your feedbacks. This latest commit addresses those issues.

src/libcore/python/xml_v.cpp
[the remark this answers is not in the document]
merlinND: I think that even if we are able to construct objects from `Properties`, all of the work would still end up in this `load_dict` pybind function because we have to instantiate the children in order to create the `props` for the parents. Right?

src/libcore/tests/test_xml.py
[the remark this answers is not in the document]
merlinND: Yes, it should be valid here too :)

include/mitsuba/core/xml.h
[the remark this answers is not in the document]
merlinND: Okay!

src/libcore/python/xml_v.cpp
```diff
@@ -83,12 +83,16 @@ std::string get_type(const py::dict &dict) {
...
-    bool within_emitter = (type == "area") || (type == "constant") ||
-                          (type == "directional") || (type == "envmap") ||
-                          (type == "point");
+    const Class *class_;
+    if (type == "scene")
+        class_ = Class::for_name("Scene", GET_VARIANT());
+    else
+        class_ = PluginManager::instance()->get_plugin_class(type, GET_VARIANT());
+
+    bool within_emitter = (class_->parent()->name() == "Emitter");
```
merlinND, line 93: I think you should go up the full hierarchy in case of more inheritance levels.

src/libcore/xml.cpp
```diff
@@ -78,6 +78,12 @@ struct Version {
     }
 };
 
+// Check if the name corresponds to a spectral IOR property which are unbounded
+// and require special handling
+bool is_ior(const std::string &name) {
```
merlinND, line 83: Maybe `is_unbounded_spectrum`? (in case there are more like those that come up)

src/libcore/python/xml_v.cpp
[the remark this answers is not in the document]
Speierers: Indeed, you would have to instanciate the children one by one manually..

src/libcore/tests/test_xml.py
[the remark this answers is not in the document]
Speierers: Ok, there is a test for this already 😉

src/libcore/python/xml_v.cpp
Speierers: The XML parser doesn't do that.. There must be a reason 🤔

Speierers:
The latest commits add:
- support for references (see the doc). This is only relevant for dictionaries of type `scene`.
- `xml.load_dict()` now also properly handles objects that need to be expanded.

src/emitters/directional.cpp
```diff
@@ -51,7 +51,7 @@ class DirectionalEmitter final : public Emitter<Float, Spectrum> {
                 Throw("Only one of the parameters 'direction' and 'to_world' "
                       "can be specified at the same time!'");
 
-            ScalarVector3f direction(normalize(props.vector3f("direction")));
+            ScalarVector3f direction(normalize(props.array3f("direction")));
```
wjakob, line 54: I would prefer it if we still have a thin wrapper `vector3f` and `point3f` that perform a cast from the `Array3f` being stored.

wjakob: Fetching an `array3f` value from the `Properties` object just looks a little strange.

src/libcore/properties.cpp
```diff
@@ -365,6 +358,70 @@ size_t Properties::size_(const std::string &name, const size_t &def_val) const {
...
+    d->entries[name].queried = false;
+}
+
+/// Array3f getter (without default)
+Array3f Properties::array3f(const std::string &name) const {
+    const auto it = d->entries.find(name);
+    if (it == d->entries.end())
+        Throw("Property \"%s\" has not been specified!", name);
+        if (!it->second.data.is<Array3f>())
+        Throw("The property \"%s\" has the wrong type (expected <array3f>).", name);
```
wjakob, line 409: Fix error message (`<array3f>` is not a supported XML tag)

src/libcore/properties.cpp
```diff
@@ -365,6 +358,70 @@ size_t Properties::size_(const std::string &name, const size_t &def_val) const {
...
+    return (Array3f) it->second.data;
+}
+
+/// Array3f getter (with default)
+Array3f Properties::array3f(const std::string &name, const Array3f &def_val) const {
+    const auto it = d->entries.find(name);
+    if (it == d->entries.end())
+        return def_val;
+        if (!it->second.data.is<Array3f>())
+        Throw("The property \"%s\" has the wrong type (expected <array3f>).", name);
```
wjakob, line 420: ditto

src/libcore/python/object.cpp
```diff
@@ -30,7 +31,15 @@ class PyTraversalCallback : public TraversalCallback {
 };
 
 MTS_PY_EXPORT(Object) {
-    py::class_<Class>(m, "Class", D(Class));
+    py::class_<Class, std::unique_ptr<Class, py::nodelete>>(m, "Class", D(Class))
```
wjakob, line 34: I don't think `std::unique_ptr<Class, py::nodelete>` is needed if we use a `py::return_value_policy` of `reference` consistently everywhere when returning classes

src/libcore/python/xml_v.cpp
```diff
@@ -38,8 +49,226 @@ MTS_PY_EXPORT(xml) {
...
+
+        SET_PROPS(py::bool_, bool, set_bool);
+        SET_PROPS(py::int_, int64_t, set_long);
+        SET_PROPS(py::float_, Properties::Float, set_float);
+        SET_PROPS(py::str, std::string, set_string);
+        SET_PROPS(Properties::Array3f, Properties::Array3f, set_array3f);
+        SET_PROPS(Properties::Transform4f, Properties::Transform4f, set_transform);
+
+        // Cast py::list and numpy.ndarray to Array3f
+        if (py::isinstance<py::list>(value) ||
```
wjakob, line 134: You can make it more general and accept any type here (even `tuple` or other classes with the right duck typing) by attempting a cast to some reasonable type (``std::array<float, 3>`` or directly ``Array3f``). Note that this may fail with a `py::cast_error`, but that is something you can catch.

wjakob: Also note the dangers of directly calling Python C API functions. For example, `PyObject_Length` can fail and return `-1` in which case, an error state is set. Continuing without resetting this state can have unintended consequences. Pybind11 wraps many python function calls and converts such errors into an exception of type `py::error_already_set` (which means that there is a C++ exception, and it's linked to a currently set Python error object). If you catch and ignore such an exception, its destructor will clear the Python error state.

src/libcore/tests/test_xml.py
```diff
@@ -297,3 +298,486 @@ def test20_upgrade_tree(variant_scalar_rgb):
                                <float name="intIOR" value="1.33"/>
                            </bsdf>
                        </scene>""")
+
+
+def test21_dict_plugins(variants_all_rgb):
```
wjakob, line 303: Let's put these into a separate file (e.g. `test_dict.py`) since it's not really related to XML, and the original test file is already quite large.

src/libcore/python/xml_v.cpp
Speierers: Great suggestion!

Speierers:
Thanks for your hard work @arpit15 I will finally merge this PR! 🚀

arpit15:
@Speierers thanks for doing all the heavy lifting. Happy to contribute to this awesome project.

PR #49, opened 2020-03-26
WIP rgb sky plugin

mjaillot:


src/emitters/sky.cpp
```diff
@@ -0,0 +1,203 @@
...
+        Float gamma = safe_acos(cos_gamma);
+
+        Spectrum result;
+        for (size_t i = 0; i < Spectrum::Size; i++) {
+            if constexpr (Spectrum::Size == 3)
+                result[i] = (Float) (arhosek_tristim_skymodel_radiance(
+                    m_state[i], (double)theta, (double)gamma, i) / 106.856980); // (sum of Spectrum::CIE_Y)
+            else {
+                Float step_size = (MTS_WAVELENGTH_MIN - MTS_WAVELENGTH_MIN) / (Float) Spectrum::Size;
+                Float wavelength0 = MTS_WAVELENGTH_MIN + step_size * i;
```
merlinND: (When the time comes to use spectral mode), the wavelengths to evaluate are already specified in `si.wavelength`.

src/emitters/sky.cpp
```diff
@@ -0,0 +1,203 @@
...
+                Float wavelength1 = MTS_WAVELENGTH_MIN + step_size * (i+1);
+                result[i] = (Float) arhosekskymodel_radiance(
+                    m_state[i], (double)theta, (double)gamma, 0.5f * (wavelength0 + wavelength1));
+            }
+        }
+
+        for (size_t i = 0; i < Spectrum::Size; i++)
+            result[i] = max(result[i], 0.0f);
```
merlinND: `result = max(result, 0.f)` should work (courtesy of Enoki).

src/emitters/sky.cpp
```diff
@@ -0,0 +1,203 @@
...
+           about the scene and default to the unit bounding sphere. */
+        m_bsphere = ScalarBoundingSphere3f(ScalarPoint3f(0.f), 1.f);
+
+        m_flags = +EmitterFlags::Infinite;
+
+        m_scale = props.float_("scale", 1.0f);
+        m_turbidity = props.float_("turbidity", 3.0f);
+        m_stretch = props.float_("stretch", 1.0f);
+        m_resolution = props.int_("resolution", 512);
+        m_albedo = props.texture<Texture>("albedo", 0.2f);
```
merlinND: I'm surprised this works: `m_albedo` has a `Spectrum` type, but here the properties you query has type `ref<Texture>`.
If you try to print the value of `albedo` here, are you getting what you expect?

I think the correct usage would be to change the type of `m_albedo` to `ref<Texture>`, and then query it like so in your `eval` method:

https://github.com/mitsuba-renderer/mitsuba2/blob/f30bd5cfe69340087597a9d684ec7b2c61718c4f/src/bsdfs/diffuse.cpp#L99

src/emitters/sky.cpp
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        Float sun_elevation = 0.5f * math::Pi<Float> - m_sun.elevation;
+
+        if (sun_elevation < 0)
+            Log(Error, "The sun is below the horizon -- this is not supported by the sky model.");
+
+        for (size_t i = 0; i < Spectrum::Size; i++) {
+            if constexpr (Spectrum::Size == 3)
+                m_state[i] = arhosek_rgb_skymodelstate_alloc_init(
```
merlinND: Since the model states are allocated here, they should also be freed in the destructor (`arhosekskymodelstate_free`, and corresponding in the spectral case).

src/emitters/sky.cpp
```diff
@@ -0,0 +1,203 @@
...
+        for (size_t i = 0; i < Spectrum::Size; i++) {
+            if constexpr (Spectrum::Size == 3)
+                result[i] = (Float) (arhosek_tristim_skymodel_radiance(
+                    m_state[i], (double)theta, (double)gamma, i) / 106.856980); // (sum of Spectrum::CIE_Y)
+            else {
+                Float step_size = (MTS_WAVELENGTH_MIN - MTS_WAVELENGTH_MIN) / (Float) Spectrum::Size;
+                Float wavelength0 = MTS_WAVELENGTH_MIN + step_size * i;
+                Float wavelength1 = MTS_WAVELENGTH_MIN + step_size * (i+1);
+                result[i] = (Float) arhosekskymodel_radiance(
+                    m_state[i], (double)theta, (double)gamma, 0.5f * (wavelength0 + wavelength1));
```
merlinND: You've probably already checked, but of course there's always the question of whether angles are expected in degrees or radians.

Speierers:
@merlinND can we close this PR?

PR #53, opened 2020-03-30
Lower CUDA Compute Capability requirement

PidgeyBE:
Together with a change in Enoki (https://github.com/mitsuba-renderer/enoki/pull/71), this allows older GPU's (tested on GeForce 940M) to use the gpu_autodiff flavors of mitsuba2.
50 is the lowest we can go as Optix 6.5 requires at least Maxwell architecture: https://en.wikipedia.org/wiki/CUDA


wjakob:
Done, thanks!

PR #60, opened 2020-04-01
Distant light source plugin

leroyvn:
This is a tentative implementation of a distant light source plugin. A test scene file is included in the `resources/test_scenes` directory and should give something like this:
test_distant

### DIFF
--- a/src/emitters/directional.cpp
+++ b/src/emitters/directional.cpp
@@ -1,0 +1,167 @@
+#include <mitsuba/core/bsphere.h>
+#include <mitsuba/core/properties.h>
+#include <mitsuba/core/warp.h>
+#include <mitsuba/render/emitter.h>
+#include <mitsuba/render/scene.h>
+#include <mitsuba/render/texture.h>
+
+NAMESPACE_BEGIN(mitsuba)
+
+/**!
+
+.. _emitter-distant:
+
+Distant directional emitter (:monosp:`directional`)
+---------------------------------------------------
+
+.. pluginparameters::
+
+    * - irradiance
+      - |spectrum|
+      - Spectral irradiance, which corresponds to the amount of spectral power
+        per unit area received by a hypothetical surface normal to the specified
+        direction.
+    * - to_world
+      - |transform|
+      - Emitter-to-world transformation matrix.
+    * - direction
+      - |vector|
+      - Alternative (and exclusive) to `to_world`. Direction towards which the
+        emitter is radiating in world coordinates.
+
+This emitter plugin implements a distant directional source which radiates a
+specified power per unit area along a fixed direction. By default, the emitter
+radiates in the direction of the positive Z axis, i.e. :math:`(0, 0, 1)`.
+
+*/
+
+template <typename Float, typename Spectrum>
+class DirectionalEmitter final : public Emitter<Float, Spectrum> {
+public:
+    MTS_IMPORT_BASE(Emitter, m_flags, m_world_transform)
+    MTS_IMPORT_TYPES(Scene, Texture)
+
+    DirectionalEmitter(const Properties &props) : Base(props) {
+        /* Until `set_scene` is called, we have no information
+           about the scene and default to the unit bounding sphere. */
+        m_bsphere = ScalarBoundingSphere3f(ScalarPoint3f(0.f), 1.f);
+
+        if (props.has_property("direction")) {
+            if (props.has_property("to_world"))
+                Throw("Only one of the parameters 'direction' and 'to_world' "
+                      "can be specified at the same time!'");
+
+            ScalarVector3f direction(normalize(props.vector3f("direction")));
+            auto [up, unused] = coordinate_system(direction);
+
+            m_world_transform =
+                new AnimatedTransform(ScalarTransform4f::look_at(
+                    ScalarPoint3f(0.0f), ScalarPoint3f(direction), up));
+        }
+
+        m_flags      = EmitterFlags::Infinite | EmitterFlags::DeltaDirection;
+        m_irradiance = props.texture<Texture>("irradiance", Texture::D65(1.f));
+    }
+
+    void set_scene(const Scene *scene) override {
+        m_bsphere = scene->bbox().bounding_sphere();
+        m_bsphere.radius =
+            max(math::RayEpsilon<Float>,
+                m_bsphere.radius * (1.f + math::RayEpsilon<Float>) );
+    }
+
+    Spectrum eval(const SurfaceInteraction3f & /*si*/,
+                  Mask /*active*/) const override {
+        return 0.f;
+    }
+
+    std::pair<Ray3f, Spectrum> sample_ray(Float time, Float wavelength_sample,
+                                          const Point2f &spatial_sample,
+                                          const Point2f & /*direction_sample*/,
+                                          Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleRay, active);
+
+        // 1. Sample spectrum
+        auto [wavelengths, weight] =
+            sample_wavelength<Float, Spectrum>(wavelength_sample);
+
+        // 2. Sample spatial component
+        const Transform4f &trafo = m_world_transform->eval(time, active);
+        Point2f p = warp::square_to_uniform_disk_concentric(spatial_sample);
+        Vector3f perp_offset = trafo.transform_affine(
+            Vector3f{ p.x(), p.y(), 0.f } * m_bsphere.radius);
+
+        // 3. Sample directional component
+        Vector3f d = trafo.transform_affine(Vector3f{ 0.f, 0.f, 1.f });
+
+        return std::make_pair(
+            Ray3f(m_bsphere.center - d * m_bsphere.radius + perp_offset, d,
+                  time, wavelengths),
+            unpolarized<Spectrum>(weight) *
+                (4.f * sqr(math::Pi<Float> * m_bsphere.radius)));
+    }
+
+    std::pair<DirectionSample3f, Spectrum>
+    sample_direction(const Interaction3f &it, const Point2f & /*sample*/,
+                     Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleDirection, active);
+
+        Vector3f d = m_world_transform->eval(it.time, active)
+                         .transform_affine(Vector3f{ 0.f, 0.f, 1.f });
+        Float dist = 2.f * m_bsphere.radius;
+
+        DirectionSample3f ds;
+        ds.p      = it.p - d * dist;
+        ds.n      = d;
+        ds.uv     = Point2f(0.f);
+        ds.time   = it.time;
+        ds.pdf    = 1.f;
+        ds.delta  = true;
+        ds.object = this;
+        ds.d      = -d;
+        ds.dist   = dist;
+
+        SurfaceInteraction3f si = zero<SurfaceInteraction3f>();
+        si.wavelengths          = it.wavelengths;
+
+        // No need to divide by the PDF here (always equal to 1.f)
+        return std::make_pair(
+            ds, unpolarized<Spectrum>(m_irradiance->eval(si, active)));
+    }
+
+    Float pdf_direction(const Interaction3f & /*it*/,
+                        const DirectionSample3f & /*ds*/,
+                        Mask /*active*/) const override {
+        return 0.f;
+    }
+
+    ScalarBoundingBox3f bbox() const override {
+        /* This emitter does not occupy any particular region
+        of space, return an invalid bounding box */
+        return ScalarBoundingBox3f();
+    }
+
+    void traverse(TraversalCallback *callback) override {
+        callback->put_object("irradiance", m_irradiance.get());
+    }
+
+    std::string to_string() const override {
+        std::ostringstream oss;
+        oss << "DirectionalEmitter[" << std::endl
+            << "  irradiance = " << string::indent(m_irradiance) << ","
+            << std::endl
+            << "  bsphere = " << m_bsphere << "," << std::endl
+            << "]";
+        return oss.str();
+    }
+
+    MTS_DECLARE_CLASS()
+
+protected:
+    ref<Texture> m_irradiance;
+    ScalarBoundingSphere3f m_bsphere;
+};
+
+MTS_IMPLEMENT_CLASS_VARIANT(DirectionalEmitter, Emitter)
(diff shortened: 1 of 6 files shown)


src/emitters/distant.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    Spectrum eval(const SurfaceInteraction3f &si, Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointEvaluate, active);
+        Vector3f v = m_world_transform->eval(si.time, active)
+                         .inverse()
+                         .transform_affine(-si.wi);
+        return (allclose(v, Vector3f{ 0.f, 0.f, 1.f }))
+                   ? unpolarized<Spectrum>(m_irradiance->eval(si, active))
+                   : unpolarized<Spectrum>(0.f);
```
Speierers, line 80: Here you should use a `select` statement as `v` might be a array of vectors (e.g. GPU variant).

Speierers: Also, this light having a `DeltaDirection`, should `eval` always return `0.f`?

src/emitters/distant.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    std::pair<Ray3f, Spectrum> sample_ray(Float /*time*/,
+                                          Float /*wavelength_sample*/,
+                                          const Point2f & /*sample2*/,
+                                          const Point2f & /*sample3*/,
+                                          Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleRay, active);
+        NotImplementedError("sample_ray");
```
Speierers, line 89: This method will be useful in the future (e.g. for a bidirectional path tracer). It would be great to have it implemented here.

src/emitters/distant.cpp
```diff
@@ -0,0 +1,157 @@
...
+        ds.delta  = true;
+        ds.object = this;
+        ds.d      = -d;
+        ds.dist   = dist;
+
+        SurfaceInteraction3f si = zero<SurfaceInteraction3f>();
+        si.wavelengths          = it.wavelengths;
+
+        return std::make_pair(
+            ds, unpolarized<Spectrum>(m_irradiance->eval(si, active)) / ds.pdf);
```
Speierers, line 116: Not necessary to divide by pdf here as it is alway `1.f`

src/emitters/distant.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    Float pdf_direction(const Interaction3f &, const DirectionSample3f &ds,
+                        Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointEvaluate, active);
+
+        Vector3f d = m_world_transform->eval(ds.time, active)
+                         .transform_affine(Vector3f{ 0.f, 0.f, 1.f });
+        return allclose(ds.d, -d) ? 1.0f : 0.0f;
+    }
```
Speierers, line 126: Should always return `0.f` as it is a `DeltaDirection` light.

src/emitters/distant.cpp
```diff
@@ -0,0 +1,157 @@
...
+            ScalarVector3f direction(normalize(props.vector3f("direction")));
+            auto [up, unused] = coordinate_system(direction);
+
+            m_world_transform =
+                new AnimatedTransform(ScalarTransform4f::look_at(
+                    ScalarPoint3f(0.0f), ScalarPoint3f(direction), up));
+        }
+
+        m_flags      = EmitterFlags::Infinite | EmitterFlags::DeltaDirection;
+        m_irradiance = props.texture<Texture>("irradiance", Texture::D65(1.f));
```
Speierers, line 63: Maybe we could change `irradiance` to `radiance` throughout this whole file for consistency with `area.cpp` and other light source?

src/emitters/distant.cpp
```diff
@@ -0,0 +1,157 @@
+#include <mitsuba/core/bsphere.h>
+#include <mitsuba/core/plugin.h>
+#include <mitsuba/core/properties.h>
+#include <mitsuba/core/warp.h>
```
Speierers, line 4: Remove unnecessary includes here

src/emitters/tests/test_distant.py
```diff
@@ -0,0 +1,119 @@
...
+         [0, 0, 0, 1]]
+    )
+
+
+@pytest.mark.parametrize("spectrum_key", spectrum_keys)
+def test_eval(variant_scalar_spectral, spectrum_key):
+    # Check correctness of the eval() method
+    from mitsuba.render import SurfaceInteraction3f
+
+    direction = np.array([0, 0, -1])
```
Speierers, line 72: please use `mitsuba.core.Vector3f` and remove `import numpy as np` in this test file

src/emitters/tests/test_distant.py
```diff
@@ -0,0 +1,119 @@
...
+    assert ek.allclose(emitter.eval(it), null_spectrum.eval(it))
+
+
+@pytest.mark.parametrize("spectrum_key", spectrum_keys)
+def test_sample_direction(variant_scalar_spectral, spectrum_key):
+    # Check correctness of sample_direction() and pdf_direction() methods
+
+    from mitsuba.render import SurfaceInteraction3f
+
+    direction = np.array([0, 0, -1])
```
Speierers, line 98: Maybe do a `for` loop over a couple directions here?

src/emitters/distant.cpp
```diff
@@ -0,0 +1,157 @@
...
+                Throw("Only one of the parameters 'direction' and 'to_world' "
+                      "can be specified at the same time!'");
+
+            ScalarVector3f direction(normalize(props.vector3f("direction")));
+            auto [up, unused] = coordinate_system(direction);
+
+            m_world_transform =
+                new AnimatedTransform(ScalarTransform4f::look_at(
+                    ScalarPoint3f(0.0f), ScalarPoint3f(direction), up));
+        }
```
Speierers, line 60: should there be an `else` statement for when only `to_world` is defined?

resources/test_scenes/test_distant.xml
```diff
@@ -0,0 +1,44 @@
+<scene version="2.0.0">
```
Speierers, line 1: This file should be moved to `resources/data/tests/scenes/somedirectory` so it will be picked by the rendergin test suite as described [here](https://mitsuba2.readthedocs.io/en/latest/src/developer_guide/testing.html#rendering-test-suite-and-z-test)

src/emitters/distant.cpp
leroyvn: Regarding the use of `select`, I'm still missing an equivalent to `allclose` which would only operate on each packet channel. This is not working as intended:
```suggestion
        return select(allclose(v, Vector3f{ 0.f, 0.f, 1.f }),
                      unpolarized<Spectrum>(m_irradiance->eval(si, active)),
                      unpolarized<Spectrum>(0.f));
```

As for the return value of `eval`, came up with this behaviour by reading the Mitsuba 1 implementation; but I honestly don't know what all the consequences could be, maybe I could miss something?

src/emitters/distant.cpp
leroyvn: I had an implementation for it in the beginning, I'll put it back here.

src/emitters/distant.cpp
leroyvn: Noted.

src/emitters/distant.cpp
leroyvn: I also got this from the Mitsuba 1 implementation; should I change it?

src/emitters/distant.cpp
leroyvn: You're right, we're actually returning a radiance (not divided by `$\pi$`). I'll fix that.

src/emitters/distant.cpp
leroyvn: Noted.

src/emitters/distant.cpp
Speierers: Mitsuba 2 doesn't have this notion of `measure`, I would say we are always in `ESolidAngle` therefore `eval` and `pdf` should always return `0.f`

leroyvn: How about that? I basically did what is in `allclose`, without the nested horizontal operation.

```suggestion
        constexpr float rtol = 1e-5f;
        constexpr float atol = 1e-8f;

        auto mask = all(abs(ScalarVector3f{ 0.f, 0.f, 1.f } - v) <=
                        abs(v) * rtol + atol);

        return select(mask, 
                      unpolarized<Spectrum>(m_radiance->eval(si, active)),
                      unpolarized<Spectrum>(0.f));
```

src/emitters/tests/test_distant.py
leroyvn: I parametrised the tests with a set of arbitrary directions.

src/emitters/distant.cpp
leroyvn: I guess so, although I actually don't really see a use case for that. I also realise that there is not default, I'll add one.

resources/test_scenes/test_distant.xml
leroyvn: Should I fork the data repo as well and create a PR there?

leroyvn:
@Speierers I pushed a couple of commits which fix some of the points you raised. I also renamed the plugin `directional` to remain consistent with Mitsuba 1 and improve scene file compatibility.

The following points remain open:

- I still have to fix the `eval` and `pdf_direction` methods (I missed your reply).
- I'm now realising that the Mitsuba 1 plugin had `irradiance` as its parameter name; I don't know what to do (consistency with `area` vs scene file comptability).
- I couldn't find out how to instantiate an identity `Transform` and I'm not happy with the workaround I used (scale with factor 1).
- The `sample_ray` method still needs care, I'll need to discuss a bit more.

src/emitters/distant.cpp
Speierers: I would still argue that this is wrong as the probability of sampling the peak of a delta distribution is zero. You can have a look at `dielectric.cpp` which is a `Delta` BSDF (same idea). There `eval` and `pdf` always return `0.f`, so it should be the same here.

src/emitters/directional.cpp
```diff
@@ -59,7 +60,8 @@ class DirectionalEmitter final : public Emitter<Float, Spectrum> {
         } else if (props.has_property("to_world")) {
             m_world_transform = props.animated_transform("to_world");
         } else {
-            m_world_transform = new AnimatedTransform(ScalarTransform4f::scale(1.));
+            m_world_transform =
```
Speierers, line 63: I just realized that the `Endpoint` constructor already initializes the `m_wolrd_transform` transform if provided. So you original code was correct.

Speierers:
Great progress!

> * I still have to fix the `eval` and `pdf_direction` methods (I missed your reply).
I just replied above.

> * I'm now realising that the Mitsuba 1 plugin had `irradiance` as its parameter name; I don't know what to do (consistency with `area` vs scene file comptability).
Okay then let's switch back to `irradiance` to be consistent with Mitsuba 1 (sorry for that)

> * I couldn't find out how to instantiate an identity `Transform` and I'm not happy with the workaround I used (scale with factor 1).
See comment in the code.

> * The `sample_ray` method still needs care, I'll need to discuss a bit more.
See comments in the code

src/emitters/directional.cpp
```diff
@@ -89,13 +91,33 @@ class DirectionalEmitter final : public Emitter<Float, Spectrum> {
...
         NotImplementedError("sample_ray");
+        // TODO: finish this
+
+        // 1. Sample spectrum
+        auto [wavelengths, weight] =
+            sample_wavelength<Float, Spectrum>(wavelength_sample);
+
+        // 2. Sample spatial component
+        // TODO: sample a point on such that cast ray will hit the scene
+        Vector3f v0 = warp::square_to_uniform_sphere(sample2);
```
Speierers, line 110: This could should be similar to the Mitsuba 1 implementation (sample on disk and transform)

src/emitters/distant.cpp
leroyvn: Ok, I fixed that.

src/emitters/distant.cpp
leroyvn: It's fixed.

src/emitters/directional.cpp
leroyvn: See latest commits for an implementation.

leroyvn:
Thanks a lot for your feedback and help. I fixed the `eval` and `pdf_position` methods, rolled back the parameter name change and the default transform argument, and proposed an implementation for the `sample_ray` method. For the latter, I essentially reproduced what's in Mitsuba 1.

I wrote the tests for the scalar RGB variant but I also tried in packet mode (just one line needs to be modified for that purpose). I'm not sure how I should check wavelengths.

On a separate note, I wonder: shouldn't be spectral sampling handled by the attached spectrum (`m_irradiance`) rather than the `sample_wavelength` template function? 

In general, I think it would be interesting to allow for more flexibility with respect to spectral sampling (I'm using Mitsuba for applications other than rendering and spectral features are critical for us). I can open a discussion on that topic if you think it's appropriate.

resources/test_scenes/test_distant.xml
Speierers: Indeed, and have a commit here that updates the submodule.

Speierers:
> In general, I think it would be interesting to allow for more flexibility with respect to spectral sampling (I'm using Mitsuba for applications other than rendering and spectral features are critical for us). I can open a discussion on that topic if you think it's appropriate.

Not sure how to answer this. Let's open an issue for that so we can discuss this futher there!

Speierers:
Everything looks good to me! 

Could you maybe squash this PR into a single commit and add the scene to the rendering test suite?

Once this is done we should be able to rebase and merge this PR to master 🚀 

Thanks a lot for your contribution!

leroyvn:
I squashed (and fixed a missing directory creation in `test_renders.py`) but haven't updated yet the data submodule. I'll wait until my PR there is accepted, then I'll update the submodule by amending this commit and we'll be ready to merge.

Speierers:
Great! Just reduce the resolution of the test render and then we will be able to merge !

Speierers:
I just merged the PR in mitsuba-data. Go ahead add the commit to update the submodule

leroyvn:
Done

Speierers:
Thanks again!

leroyvn:
Thank you as well for the review and assistance!

PR #64, opened 2020-04-04
Port `ref_vector` from Mitsuba1

pjessesco:


Speierers:
Hi @pjessesco ,

Could you maybe give us a little bit of context regarding this PR? I am not sure to see where this `ref_vector` would be used in the codebase?

pjessesco:
Hi,

I'm porting a custom plugin that is implemented in Mitsuba 1(that uses `ref_vector`), and I found that it is replaced to `std::vector<>` in Mitsuba 2.

For example, in `scene.h`,

Mitsuba 2

    ...
    std::vector<ref<Shape>> m_shapes;
    std::vector<ref<Sensor>> m_sensors;
    std::vector<ref<Object>> m_children;
    ref<Integrator> m_integrator;
    ...

Mitsuba 1

	...
    ref_vector<Shape> m_shapes;
    ref_vector<Shape> m_specialShapes;
    ref_vector<Sensor> m_sensors;
    ref_vector<Emitter> m_emitters;
    ref_vector<ConfigurableObject> m_objects;
    ...

Speierers:
Could your custom plugin use `std::vector<>` instead of `ref_vector`?

I don't really see how other people could benefit from `ref_vector` and we would like to keep the core library as light/clean as possible. For this reason I would rather not accept this PR and let you re-implement the `ensureUnique` and the other method directly in your custom plugin if you really need it.

If you still believe other people could benefit from this, please illustrate it by a concrete example and I would be happy to discuss this further 😉

wjakob:
Mitsuba's `ref_vector` goes back to ancient times. I believe at some point there was some issue with non-POD types (like `ref<>`) in an ordinary vector in some STL versions, requiring adaptations that were later dropped (but the type persisted.) In any case, better not to revive this class again.

Issue #82, opened 2020-04-14
Results vary vs sample count while they should not


leroyvn:
When rendering a simple scene with the path tracer, my results vary with the number of samples. The scene is composed of a square surface with the default `diffuse` BSDF illuminated by an environment emitter (`constant` or `directional`) (see below a render at 32 SPP with the constant emitter).

![scene](https://user-images.githubusercontent.com/34740232/79237502-828cf680-7e6e-11ea-8fd0-bb998307da36.png)

The code I used is on [this gist](
https://gist.github.com/leroyvn/1f6dac326908f7c482d1570e3cf4072f). I'm using a perspective camera with a single pixel and a very small field of view to make sure that I only see a small part of the square and no background. With both emitters, I vary the number of SPP by powers of 10 up to 1e8. The theoretical solution is `rho` (the surface's reflectance) for the `constant` emitter case and `rho / pi` for the `directional` emitter case. Results are as follows:

```
emitter      spp    li
------------ ------ -----------
constant     1e+00  0.256591797
constant     1e+01  0.50390625
constant     1e+02  0.484130859
constant     1e+03  0.492675781
constant     1e+04  0.499267578
constant     1e+05  0.5
constant     1e+06  0.500488281
constant     1e+07  0.505371094
constant     1e+08  1.0
------------ ------ -----------
directional  1e+00  0.159179688
directional  1e+01  0.159179688
directional  1e+02  0.159179688
directional  1e+03  0.159179688
directional  1e+04  0.159179688
directional  1e+05  0.159179688
directional  1e+06  0.157592773
directional  1e+07  0.153076172
directional  1e+08  0.25
```
Or, as a plot:
![results_vs_spp](https://user-images.githubusercontent.com/34740232/79237165-24f8aa00-7e6e-11ea-9c85-b05b320d0196.png)

Depending on the number of samples taken, the incoming radiance at the sensor varies. I'm surprised to see something like that since the value returned by either emitter is constant. I'm actually also surprised to see that the results plateau with good accuracy until 100.000 samples for the `directional` case and do not for the `constant` case. Discrepancies become high when reaching 100 million SPP.

Any idea of what could be the cause of that?

# System configuration

Platform: macOS 10.15.4
Compiler: Apple clang version 11.0.0 (clang-1100.0.33.17)
Python version: Python 3.7.5 :: Anaconda, Inc.
Mitsuba commit hash: bad308d4a993721a4c89819e362d0fa0e7aa46d3
[...]

PR #83, opened 2020-04-14
Radiancemeter sensor plugin

leroyvn:
This commit ports Mitsuba 1's radiance meter sensor.

Co-authored-by: @schunkes

### DIFF
--- a/src/sensors/radiancemeter.cpp
+++ b/src/sensors/radiancemeter.cpp
@@ -1,0 +1,153 @@
+#include <mitsuba/core/fwd.h>
+#include <mitsuba/core/properties.h>
+#include <mitsuba/core/transform.h>
+#include <mitsuba/render/fwd.h>
+#include <mitsuba/render/sensor.h>
+
+NAMESPACE_BEGIN(mitsuba)
+
+/**!
+
+.. _sensor-radiancemeter:
+
+Radiance meter (:monosp:`radiancemeter`)
+----------------------------------------
+
+.. pluginparameters::
+
+ * - to_world
+   - |transform|
+   - Specifies an optional camera-to-world transformation.
+     (Default: none (i.e. camera space = world space))
+ * - origin
+   - |point|
+   - Location from which the sensor will be recording in world coordinates.
+     Must be used with `origin`.
+ * - direction
+   - |vector|
+   - Alternative (and exclusive) to `to_world`. Direction in which the
+     sensor is pointing in world coordinates. Must be used with `origin`.
+
+This sensor plugin implements a simple radiance meter, which measures
+the incident power per unit area per unit solid angle along a
+certain ray. It can be thought of as the limit of a standard
+perspective camera as its field of view tends to zero.
+This sensor is used with films of 1 by 1 pixels.
+
+Such a sensor is useful for conducting virtual experiments and
+testing the renderer for correctness.
+
+By default, the sensor is located at the origin and performs
+a measurement in the positive Z direction :monosp:`(0,0,1)`. This can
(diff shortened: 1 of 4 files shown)


src/sensors/radiancemeter.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+        // 2. Set ray origin and direction
+        auto trafo = m_world_transform->eval(time, active);
+        ray.o      = trafo.transform_affine(Point3f{ 0.f, 0.f, 0.f });
+        ray.d      = trafo.transform_affine(Vector3f{ 0.f, 0.f, 1.f });
+
+        // 3. Set differentials
+        ray.d_x               = 0.f;
+        ray.d_y               = 0.f;
+        ray.has_differentials = true;
```
Speierers, line 125: Not sure whether we should even set the derivatives here. 
As this sensor operates on a single pixel it makes sense to that `d_x, d_y` should be equal to zero.

Maybe we could simply leave those with their default values (`ray.has_differentials` will be `false`)?

Speierers:
Hi @leroyvn, 

Thanks a lot for opening this PR. Really happy to see people porting/implementing more plugins in Mitsuba 2!

The code looks great to me, and the tests as well. Just one minor comment (see above).

I am curious to know if you have encountered any bugs when rendering 1x1 images with this plugin. Not sure this is something we have ever tested.

leroyvn:
I had similar behaviour as in #82 with both the `perspective` and `radiancemeter` plugins, but I don't think it's due to the film size. 

Forcing the use of a `box` reconstruction filter could be good. Maybe I should change something to implement this behaviour?

Speierers:
> Forcing the use of a box reconstruction filter could be good. Maybe I should change something to implement this behaviour?

Indeed it would be nice to have a warning in the constructor if the `film->reconstruction_filter()` isn't a box filter. Or maybe it is enough to check whether the `radius` of that filter is `<= 0.5f`?

One other thing: could we add a test where we construct a simple scence (e.g. no shapes, just a `constant` emitter) with this sensor, and call `scene.integrator().render()`?

leroyvn:
Sure, I will do do that.

leroyvn:
There it is! I used numpy to check the results, is that okay?

src/sensors/radiancemeter.cpp
```diff
@@ -77,6 +77,11 @@ MTS_VARIANT class RadianceMeter final : public Sensor<Float, Spectrum> {
 
         if (m_film->size() != ScalarPoint2i(1, 1))
             Throw("This sensor only supports films of size 1x1 Pixels!");
+
+        if (m_film->reconstruction_filter()->radius() >
+            0.5f + math::RayEpsilon<Float>)
```
Speierers, line 82: Not sure the `math::RayEpsilon` is necessary here. Am I missing someting?

I would just do the following comparison:
```
if (m_film->reconstruction_filter()->radius() > 0.5f)
```

Speierers:
> There it is! I used numpy to check the results, is that okay?

Yes, `numpy` is fine when dealing with `Bitmap`.

One minor comment regarding the filter radius check (see above) and it would be great to squash those commit into one afterthat.

Then this should be ready to launch! 🚀 

Thanks again @leroyvn

src/sensors/radiancemeter.cpp
leroyvn: I initially didn't put the extra `math::RayEpsilon`, but then I was getting the warning when using the box filter with its default settings; hence this fix.

Speierers: My bad, I didn't know `math::RayEpsilon` was used in `box.cpp`. All good then!

PR #85, opened 2020-04-15
Mitsuba 1 compatibility & Sunsky

tszirr:
This pull request includes several patches which ultimately allow to load all the mitsuba1 scenes that I have been using in the past years, mostly without any manual modifications.
I kept the patch as small as I could, the biggest additon being an as direct as possible port of the mitsuba1 sunsky plugin.

I tested scalar_rgb, packet_rgb, gpu_rgb, and scalar_spectral modes, they all approximately match the references I have for mitsuba 1.

While I tried to keep changes small, the following outline hopefully simplifies their review:
 * The first three commits are somewhat unrelated bugfixes (https://github.com/tszirr/mitsuba2-patches/commits/bugfixes), fixing XML object expansion, generic env map classification, and a utility function signature
 * The fourth commit (https://github.com/tszirr/mitsuba2-patches/commits/compatibility) brings minor improvements for compatibility with previous scene file versions: It prevents breaking camelCase conversion of <default> names, mangles names that collide with reserved IDs,  and fixes 'diffuseReflectance' for diffuse BSDFs
* The fifth commit (https://github.com/tszirr/mitsuba2-patches/commits/upgrade) adds two command line flags (-D__load_permissive=1 and -D__update_approximate=1) to the XML loader, which allow to ignore 'unreferenced X' errors and approximately replace currently unsupported BSDFs, respectively
* The final commit (https://github.com/tszirr/mitsuba2-patches/commits/sunsky) adds the mitsuba1 sunsky plugin in a slighly compacted form (merging sun, sky, and sunsky, with corresponding XML upgrade paths). The code closely follows mitsuba1, its major open issue being that there is currently no way to pass a spectral texture to the 'envmap' plugin. This results in unnecessary RGB and subsequent upsampling conversions for the spectral code path. However, I tried to keep the code forward-compatibile, such that the necessary changes to the spectral code path are minimal, once spectral emitter textures are supported.

With this set of changes (and the additional upgrade load flags), I can load and render all mitsuba1 scenes I have on drive, including many scenes from previous work and popular public scene archives.

PR #86, opened 2020-04-15
Bugfixes (XML multi-object expand(), environment map classification)

tszirr:
Bugfix to XML multi-object expansion, environment map emitter classification, and a spectrum utility signature. 

These fixes are included in the 'Compatibility & Sunsky' pull request (https://github.com/mitsuba-renderer/mitsuba2/pull/85), but separated here, since they are topically unrelated, small, and in part significant.

Description of the environment map issue: Directional lights are also marked Infinite, therefore erroneously classified as environment emitter and conflicting with other environment maps.

### DIFF
--- a/include/mitsuba/core/spectrum.h
+++ b/include/mitsuba/core/spectrum.h
@@ -344,8 +344,8 @@ inline void spectrum_from_file(const std::string &filename,
 }
 
 template <typename Scalar>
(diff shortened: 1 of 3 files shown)


Speierers:
Thanks @tszirr for those fixes, I will merge this PR.

PR #87, opened 2020-04-15
Irradiancemeter sensor plugin

schunkes:
This pull request ports Mitsuba 1's irradiance meter sensor.

Co-authored by: @leroyvn

### DIFF
--- a/src/sensors/irradiancemeter.cpp
+++ b/src/sensors/irradiancemeter.cpp
@@ -1,0 +1,123 @@
+#include <mitsuba/core/fwd.h>
+#include <mitsuba/core/properties.h>
+#include <mitsuba/core/transform.h>
+#include <mitsuba/core/warp.h>
+#include <mitsuba/render/fwd.h>
+#include <mitsuba/render/sensor.h>
+
+NAMESPACE_BEGIN(mitsuba)
+
+/**!
+
+.. _sensor-irradiancemeter:
+
+Irradiance meter (:monosp:`irradiancemeter`)
+--------------------------------------------
+
+.. pluginparameters::
+
+ * - none
+
+This sensor plugin implements an irradiance meter, which measures
+the incident power per unit area over a shape which it is attached to.
+This sensor is used with films of 1 by 1 pixels.
+
+If the irradiance meter is attached to a mesh-type shape, it will measure the
+irradiance over all triangles in the mesh.
+
+This sensor is not instantiated on its own but must be defined as a child
+object to a shape in a scene. To create an irradiance meter, 
+simply instantiate the desired sensor shape and specify an 
+:monosp:`irradiancemeter` instance as its child:
+
+.. code-block:: xml
+    :name: sphere-meter
+
+    <shape type="sphere">
+        <sensor type="irradiancemeter">
+            <!-- film -->
+        </sensor>
+    </shape>
+*/
+
+MTS_VARIANT class IrradianceMeter final : public Sensor<Float, Spectrum> {
+public:
+    MTS_IMPORT_BASE(Sensor, m_film, m_world_transform, m_shape)
+    MTS_IMPORT_TYPES(Shape)
+
+    IrradianceMeter(const Properties &props) : Base(props) {
+        if (props.has_property("to_world"))
+            Throw("Found a 'to_world' transformation -- this is not allowed. "
+                  "The irradiance meter inherits this transformation from its parent "
+                  "shape.");
+        
+        if (m_film->size() != ScalarPoint2i(1, 1))
+            Throw("This sensor only supports films of size 1x1 Pixels!");
+
+        if (m_film->reconstruction_filter()->radius() >
+            0.5f + math::RayEpsilon<Float>)
+            Log(Warn, "This sensor should only be used with a reconstruction filter"
+               "of radius 0.5 or lower(e.g. default box)");
+    }
+
+    void set_shape(Shape *shape) override {
+        if (m_shape)
+            Throw("An irradiance meter can be only be attached to a single shape.");
+
+        Base::set_shape(shape);
+    }
+
+    std::pair<RayDifferential3f, Spectrum>
+    sample_ray_differential(Float time, Float wavelength_sample,
+                            const Point2f & sample2,
+                            const Point2f & sample3,
+                            Mask active) const override {
+
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleRay, active);
+
+        // 1. Sample spatial component
+        PositionSample3f ps = m_shape->sample_position(time, sample2, active);
+
+        // 2. Sample directional component
+        Vector3f local = warp::square_to_cosine_hemisphere(sample3);
+
+        // 3. Sample spectrum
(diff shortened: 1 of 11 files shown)


src/sensors/irradiancemeter.cpp
```diff
@@ -0,0 +1,118 @@
...
+    MTS_IMPORT_TYPES(Shape)
+
+    IrradianceMeter(const Properties &props) : Base(props) {
+        if (props.has_property("to_world"))
+            Throw("Found a 'to_world' transformation -- this is not allowed. "
+                  "The irradiance meter inherits this transformation from its parent "
+                  "shape.");
+        
+        if (m_film->size() != ScalarPoint2i(1, 1))
+            Throw("This sensor only supports films of size 1x1 Pixels!");
```
Speierers, line 55: Similarly to PR #83, we should check that `filter-radius() < 0.5f` here.

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
+import math
```
Speierers, line 1: Remove this import as it is not needed.

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
...
+    </shape>
+    """
+    return load_string(xml)
+
+
+def test_construct(variant_scalar_rgb):
+    """
+    We construct an irradiance meter attached to a sphere and assert that the
+    following parameters get set correctly:
+    - to_world transform
```
Speierers, line 35: Apparently we don't check for `to_world`?

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
+import math
+import random
```
Speierers, line 2: use `np.random.rand()` instead.

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    We construct an irradiance meter attached to a sphere and assert that sampled
+    rays originate at the sphere's surface
+    """
+    from mitsuba.core import Vector3f
+
+    radius = 45.2328
+    x = 2
+    y = 5
+    z = 8.3
```
Speierers, line 60: It would be great to parameterize this on a couple position and radii using the pytest `parameterize` feature:
```python
@pytest.mark.parametrize("center", [[0.0, 0.0, 0.0], [-1.0, -1.0, 0.5], [4.0, 1.0, 0.0]])
...
```

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
...
+
+        o_rel = Vector3f(ray.o.x - x, ray.o.y - y, ray.o.z - z)
+
+        d_rel = Vector3f(o_rel + ray.d)
+
+        # assert that the ray starts at the sphere surface
+        assert ek.abs(ek.norm(o_rel) - radius) < 1e-5
+
+        # assert that all rays point away from the sphere center
+        ek.norm(d_rel) > ek.norm(o_rel)
```
Speierers, line 76: Not really sure about those tests. I would suggest the following:
```python
# assert that the ray starts at the sphere surface
assert ek.allclose(ek.norm(center - ray.o), radius)

# assert that all rays point away from the sphere center
assert ek.dot(ek.normalize(center - ray.o), ray.d) > 0.0
```

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
...
+            </film>
+        </sensor>
+    </shape>
+    """
+
+    emitter_xml = f"""
+    <emitter type="constant">
+        <spectrum name="radiance" type='uniform'>
+        <float name="value" value="{radiance}"/>
+    </spectrum>
```
Speierers, line 114: indentation typo 😉

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    scene = load_string(scene_xml)
+    sensor = scene.sensors()[0]
+
+    power_density_cum = 0.0
+    num_samples = 100
+    random.seed(42)
+
+    for i in range(num_samples):
+        r = [random.random() for x in range(6)]
```
Speierers, line 132: use `numpy` instead

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -0,0 +1,140 @@
...
+        d_rel = Vector3f(o_rel + ray.d)
+
+        # assert that the ray starts at the sphere surface
+        assert ek.abs(ek.norm(o_rel) - radius) < 1e-5
+
+        # assert that all rays point away from the sphere center
+        ek.norm(d_rel) > ek.norm(o_rel)
+
+
+def test_incoming_flux(variant_scalar_rgb):
```
Speierers, line 79: This is a great test!

Could we have the same test but this time using an actual integrator, and calling the `scene.integrator().render()`?

schunkes: I deliberately designed this test without an integrator, because I wanted it to depend on as little other parts of Mitsuba as possible.
I added another test case that does the same but uses the path tracer.

src/sensors/tests/test_irradiancemeter.py
schunkes: I implemented your suggestion, but I believe the second assertion should be 
`(ray.o - center)`

schunkes:
Thank you so much for this very detailed and instructive feedback!
I closed some discussions without further comment, since I had nothing special to add, except that I implemented your suggestion.

src/sensors/tests/test_irradiancemeter.py
```diff
@@ -126,10 +110,9 @@ def test_incoming_flux(variant_scalar_rgb):
 
     power_density_cum = 0.0
     num_samples = 100
-    random.seed(42)
 
     for i in range(num_samples):
-        r = [random.random() for x in range(6)]
+        r = [np.random.rand() for x in range(6)]
```
Speierers: I would suggest the following:

```python
wav_samples = np.random.rand(num_samples)
pos_samples = np.random.rand(num_samples, 2)
dir_samples = np.random.rand(num_samples, 2)
for i in range(num_samples):
    ray, weight = sensor.sample_ray_differential(0.0, wav_samples[i], pos_samples[i], dir_samples[i])
```

And same above

schunkes: I suppose this is more efficient, because the RNG is calles only three times, instead of 100?
I pushed some changes.

Speierers: It is probably more efficient, but my suggestion was more about keeping the code clean 😉

Speierers:
Please resolve the conflict and squash those commits into a single commit.

Then I will be happy to rebase and merge this PR! 🚀 

Thanks a lot for your contribution!

schunkes:
Everything done.
Thank you again for being so helpful!

And a nice weekend to everyone in your group! :)

PR #89, opened 2020-04-16
Bugfixes / Mitsuba 1 Compatibility

tszirr:
This brings minor fixes for compatibility with previous scene file versions: It prevents the auto camelCase conversion from breaking <default> names, mangles names that collide with reserved IDs, and fixes 'diffuseReflectance' for diffuse BSDFs.

(Cannot split the commit tree right now, hope GitHub notices that part of it was already merged. Otherwise will have to repost with all changes split up later/tomorrow)


tszirr:
Ok, apparently it does not notice the previously merged changes. Then postponed to later ...

PR #90, opened 2020-04-16
 fix <default> upgrade & improve backwards compatibility

tszirr:
This brings minor fixes for compatibility with previous scene file versions: It prevents the auto camelCase conversion from breaking \<default\> names, mangles names that collide with reserved IDs, and fixes 'diffuseReflectance' for diffuse BSDFs.

Issue #22, opened 2020-03-08
About shape groups/assemblies


KelSolaar:
Hi,

I might have missed the obvious but I could not find a way to make `shape` groups or assemblies.

In the past, with Mitsuba 1, I have been relying on translators to export data for me so it never has been an issue. I am however currently building scenes manually so I'm wondering if it is possible to create groups of shapes and transform them as a single entity. I would like, for example, assemble a colour checker with the frame and the patches and put it in place wherever I need.

Hope that makes sense,

Cheers,

Thomas

chaosink:
Maybe you can try my script: [rm_instance](https://github.com/chaosink/mitsuba_tools/tree/master/rm_instance).
It can convert `instance` and `shapegroup` shapes into non-instanced shapes.

wjakob:
Hi @KelSolaar,

instancing is one of the features that we haven't yet ported to Mitsuba 2. We are currently looking  into this, as it is obviously an important feature.

Best,
Wenzel

KelSolaar:
Sweet thanks @wjakob!

@chaosink : I searched for `group` in the entire codebase and could not find anything, hence my thread. Bookmarked your repo for future reference though!

Cheers,

Thomas

chaosink:
@KelSolaar Try searching `shapegroup` instead.

KelSolaar:
Unless I missed something super obvious (and as I stated it earlier), nothing like that in Mitsuba 2!

![image](https://user-images.githubusercontent.com/99779/76169471-60ec7f80-61dd-11ea-9db6-235cef305663.png)

chaosink:
@KelSolaar  Convert `instance` and `shapegroup` shapes in XML files to non-instanced shapes.
Is this what you want?
Or I misunderstood you needs?

KelSolaar:
I need to be able to take a bunch of shapes, group them and transform that group.

For example, I have that ColorChecker geometry here: https://github.com/colour-science/colour-mitsuba/tree/master/colour_mitsuba/colorchecker_classic/geometry. I want to assemble it a the origin of the world, group everything under a group/assembly and then move that group at the location I wish. At the moment I have to export everything at the final worldspace location, e.g. against the Cornell Box wall for that precise ColorChecker geometry.

Hope that makes sense!

chaosink:
I mean, you can first generate a XML scene file for Mitsuba1 just as before.
[...]

PR #91, opened 2020-04-16
Approximate / force upgrade modes for Mitsuba 1 files

tszirr:
Adds two command line flags (-D__load_permissive=1 and -D__upgrade_approximate=1) to the XML loader, which allow to ignore 'unreferenced X' errors and approximately replace currently unsupported features, respectively.

This PR still subsumes https://github.com/mitsuba-renderer/mitsuba2/pull/90, since however I split the changes, they will create merge conflicts, if merged simultaneously. Therefore, to merge both, the simplest option is to only merge this one and close the other one.

The **approximate load** (-D__upgrade_approximate=1) mode can do the following:
* Transform phong BSDFs and phong NDFs to beckmann w/ approx eq. alpha
* Replace binary mixture BSDFs
* Create proxies for coatings
* Bypass unsupported bumpmap / normalmap nestings
* Unroll shapegroups / instances (https://github.com/mitsuba-renderer/mitsuba2/issues/22)
* Try to convert spectrums w/o wavelength specifications to RGB

Additionally, standard upgrades are extended by:
* Converting hex color codes
* Bump -> bumpmap (which approx load then bypasses until a plugin is available)
* diffuse::diffuseReflectance -> diffuse:reflectance

All these features were required by at least one actual Mitsuba 1 scene file on the internet.

For the sunsky required by many scenes, see the old updated PR (https://github.com/mitsuba-renderer/mitsuba2/pull/85).

### DIFF
--- a/src/libcore/xml.cpp
+++ b/src/libcore/xml.cpp
@@ -424,6 +424,45 @@ void upgrade_tree(XMLSource &src, pugi::xml_node &node, const Version &version, 
                     wn.set_name("float");
                 }
             }
+            for (pugi::xpath_node result: node.select_nodes("//shape[@type='instance']")) {
+                pugi::xml_node n = result.node();
+                Log(Warn, "Unrolling instance -> shapes! \"%s\"", n.attribute("id").value());
+                int shapeCtr = 0;
+                for (pugi::xpath_node result: n.select_nodes("ref/@id")) {
+                    char const* shapegroup_id = result.attribute().value();
+                    std::string group_selector = std::string("//shape[@id='") + shapegroup_id + "' and @type='shapegroup']";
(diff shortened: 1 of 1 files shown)


wjakob:
Sorry, I'm not a big fan of this change. Mitsuba 2 aims for compatibility with Mitsuba 1, but it is still under heavy development, so compatibility is subject to a "best-effort with what is already implemented" policy.  I wouldn't want to add flags like__upgrade_approximate just to get a Mitsuba 1 scene to parse. Of course it's fine to have these types of modifications, but I don't think they should go into the master branch.

wjakob:
As always, I am happy to discuss further, but I'll close this PR for now.

PR #99, opened 2020-04-21
Spectrum: move two functions to .cpp

merlinND:
The goal is to reduce the size of `core/spectrum.h` (including the headers it includes) because it's used everywhere.

The two functions that were moved to `libcore/spectrum.cpp` were templated over `Scalar`, so they are explicitly instantiated for `float` and `double`.

Tested on Fedora 29 with modes "scalar_rgb", "scalar_spectral", "gpu_autodiff_rgb".

@Speierers, could you please check this doesn't break the Windows build?

### DIFF
--- a/src/libcore/spectrum.cpp
+++ b/src/libcore/spectrum.cpp
@@ -1,6 +1,106 @@
+#include <mitsuba/core/fresolver.h>
+#include <mitsuba/core/fstream.h>
 #include <mitsuba/core/spectrum.h>
 
 NAMESPACE_BEGIN(mitsuba)
+
+
+template <typename Scalar>
+void spectrum_from_file(const std::string &filename, std::vector<Scalar> &wavelengths,
+                        std::vector<Scalar> &values) {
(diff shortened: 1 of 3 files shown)


src/libcore/spectrum.cpp
```diff
@@ -1,7 +1,105 @@
...
+
+
+/// Explicit instantiations
+template void spectrum_from_file(const std::string &filename, std::vector<float> &wavelengths,
+                                 std::vector<float> &values);
+template void spectrum_from_file(const std::string &filename, std::vector<double> &wavelengths,
+                                 std::vector<double> &values);
+
+template Color<float, 3> spectrum_to_rgb(const std::vector<float> &wavelengths,
+                                        const std::vector<float> &values, bool bounded);
```
Speierers, line 97: Indenation typo 😋

merlinND:
Pushed the fixes we discussed for Windows.

Speierers:
Thanks @merlinND!

PR #100, opened 2020-04-22
Port of spot light from Mitsuba 1

mehrab2603:
This is an initial port of spotlight from Mitsuba 1. The implementation of sample_ray function is probably wrong as I am not sure what to return and how to calculate it. Would be glad to discuss.

Here's a render using a spot light. Tested on scalar_rgb, gpu_autodiff_spectral, gpu_autodiff_rgb.

output

### DIFF
--- a/src/emitters/spot.cpp
+++ b/src/emitters/spot.cpp
@@ -1,0 +1,198 @@
+#include <mitsuba/core/properties.h>
+#include <mitsuba/core/warp.h>
+#include <mitsuba/render/emitter.h>
+#include <mitsuba/render/medium.h>
+#include <mitsuba/render/texture.h>
+
+NAMESPACE_BEGIN(mitsuba)
+
+/**!
+
+.. _emitter-spot:
+
+Spot light source (:monosp:`spot`)
+------------------------------------
+
+.. pluginparameters::
+
+ * - intensity
+   - |spectrum|
+   - Specifies the maximum radiant intensity at the center in units of power per unit steradian. (Default: 1).
+     This cannot be spatially varying (e.g. have bitmap as type).
+ * - cutoff_angle
+   - |float|
+   - Cutoff angle, beyond which the spot light is completely black (Default: 20 degrees)
+ * - beam_width
+   - |float|
+   - Subtended angle of the central beam portion (Default: :math:`cutoff_angle \times 3/4`)
+ * - texture
+   - |texture|
+   - An optional texture to be projected along the spot light. This must be spatially varying (e.g. have bitmap as type).
+ * - to_world
+   - |transform|
+   - Specifies an optional emitter-to-world transformation.  (Default: none, i.e. emitter space = world space)
+
+This plugin provides a spot light with a linear falloff. In its local coordinate system, the spot light is
+positioned at the origin and points along the positive Z direction. It can be conveniently reoriented
+using the lookat tag, e.g.:
+
+.. code-block:: xml
+    :name: spot-light
+
+    <emitter type="spot">
+        <transform name="to_world">
+            <!-- Orient the light so that points from (1, 1, 1) towards (1, 2, 1) -->
+            <lookat origin="1, 1, 1" target="1, 2, 1"/>
+        </transform>
+    </emitter>
+
+The intensity linearly ramps up from cutoff_angle to beam_width (both specified in degrees),
+after which it remains at the maximum value. A projection texture may optionally be supplied.
+
+.. subfigstart::
+.. subfigure:: ../../resources/data/docs/images/render/emitter_spot_no_texture.jpg
+   :caption: Two spot lights with different colors and no texture specified.
+.. subfigure:: ../../resources/data/docs/images/render/emitter_spot_texture.jpg
+   :caption: A spot light with a texture specified.
+.. subfigend::
+   :label: fig-spot-light
+
+ */
+
+template <typename Float, typename Spectrum>
+class SpotLight final : public Emitter<Float, Spectrum> {
+public:
+    MTS_IMPORT_BASE(Emitter, m_flags, m_medium, m_world_transform)
+    MTS_IMPORT_TYPES(Scene, Texture)
+
+    SpotLight(const Properties &props) : Base(props) {
+        m_flags = +EmitterFlags::DeltaPosition;
+        m_intensity = props.texture<Texture>("intensity", Texture::D65(1.f));
+        m_texture = props.texture<Texture>("texture", Texture::D65(1.f));
+
+        if (m_intensity->is_spatially_varying())
+            Throw("The parameter 'intensity' cannot be spatially varying (e.g. bitmap type)!");
+
+        if (props.has_property("texture")) {
+            if (!m_texture->is_spatially_varying())
+                Throw("The parameter 'texture' must be spatially varying (e.g. bitmap type)!");
+            m_flags |= +EmitterFlags::SpatiallyVarying;
+        }
+
+        m_cutoff_angle = props.float_("cutoff_angle", 20.0f);
+        m_beam_width = props.float_("beam_width", m_cutoff_angle * 3.0f / 4.0f);
+        m_cutoff_angle = deg_to_rad(m_cutoff_angle);
+        m_beam_width = deg_to_rad(m_beam_width);
+        m_inv_transition_width = 1.0f / (m_cutoff_angle - m_beam_width);
+        m_cos_cutoff_angle = cos(m_cutoff_angle);
+        m_cos_beam_width = cos(m_beam_width);
+        Assert(m_cutoff_angle >= m_beam_width);
+        m_uv_factor = tan(m_cutoff_angle);
+    }
+
+    inline Spectrum falloff_curve(const Vector3f &d, Wavelength wavelengths, Mask active) const {
+        SurfaceInteraction3f si = zero<SurfaceInteraction3f>();
+        si.wavelengths = wavelengths;
+        Spectrum result = m_intensity->eval(si, active);
+
+        Vector3f local_dir = normalize(d);
+        const Float cos_theta = local_dir.z();
+
+        if (m_texture->is_spatially_varying()) {
+            si.uv = Point2f(0.5f + 0.5f * local_dir.x() / (local_dir.z() * m_uv_factor),
+                            0.5f + 0.5f * local_dir.y() / (local_dir.z() * m_uv_factor));
+            result *= m_texture->eval(si, active);
+        }
+
+        auto beam_res = select(cos_theta >= m_cos_beam_width, result,
+                               result * ((m_cutoff_angle - acos(cos_theta)) * m_inv_transition_width));
+
+        return select(cos_theta <= m_cos_cutoff_angle, Spectrum(0.0f), beam_res);
+    }
+
+    std::pair<Ray3f, Spectrum> sample_ray(Float time, Float wavelength_sample,
+                                          const Point2f &spatial_sample,
+                                          const Point2f & /*dir_sample*/,
+                                          Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleRay, active);
+
+        // 1. Sample directional component
+        const Transform4f &trafo = m_world_transform->eval(time, active);
+        Vector3f local_dir = warp::square_to_uniform_cone(spatial_sample, (Float)m_cos_cutoff_angle);
+        Float pdf_dir = warp::square_to_uniform_cone_pdf(local_dir, (Float)m_cos_cutoff_angle);
+
+        // 2. Sample spectrum
+        auto [wavelengths, spec_weight] = m_intensity->sample(
+            zero<SurfaceInteraction3f>(),
+            math::sample_shifted<Wavelength>(wavelength_sample), active);
+        auto falloff_spec = falloff_curve(local_dir, wavelengths, active);
+
+        return { Ray3f(trafo.translation(), trafo * local_dir, time, wavelengths),
+                unpolarized<Spectrum>(falloff_spec / pdf_dir) };
+    }
+
+    std::pair<DirectionSample3f, Spectrum> sample_direction(const Interaction3f &it,
+                                                            const Point2f &/*sample*/,
+                                                            Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleDirection, active);
+
+        auto trafo = m_world_transform->eval(it.time, active);
+
+        DirectionSample3f ds;
+        ds.p        = trafo.translation();
+        ds.n        = 0.f;
+        ds.uv       = 0.f;
+        ds.pdf      = 1.0f;
+        ds.time     = it.time;
+        ds.delta    = true;
+        ds.object   = this;
+        ds.d        = ds.p - it.p;
+        ds.dist     = norm(ds.d);
+        Float inv_dist = rcp(ds.dist);
+        ds.d        *= inv_dist;
+        Vector3f local_d = trafo.inverse() * -ds.d;
+        Spectrum falloff_spec = falloff_curve(local_d, it.wavelengths, active);
+
+        return { ds, unpolarized<Spectrum>(falloff_spec * (inv_dist * inv_dist)) };
+    }
+
+    Float pdf_direction(const Interaction3f &, const DirectionSample3f &, Mask) const override {
+        return 0.f;
+    }
+
+    Spectrum eval(const SurfaceInteraction3f &, Mask) const override { return 0.f; }
+
+    ScalarBoundingBox3f bbox() const override {
+        return m_world_transform->translation_bounds();
+    }
+
+    void traverse(TraversalCallback *callback) override {
+        callback->put_object("intensity", m_intensity.get());
+        callback->put_object("texture", m_texture.get());
+    }
+
+    std::string to_string() const override {
+        std::ostringstream oss;
+        oss << "SpotLight[" << std::endl
+            << "  world_transform = " << string::indent(m_world_transform->to_string()) << "," << std::endl
+            << "  intensity = " << m_intensity << "," << std::endl
+            << "  cutoff_angle = " << m_cutoff_angle << "," << std::endl
+            << "  beam_width = " << m_beam_width << "," << std::endl
+            << "  texture = " << (m_texture ? string::indent(m_texture->to_string()) : "")
+            << "  medium = " << (m_medium ? string::indent(m_medium->to_string()) : "")
+            << "]";
+        return oss.str();
+    }
+
+    MTS_DECLARE_CLASS()
+private:
+    ref<Texture> m_intensity;
+    ref<Texture> m_texture;
+    ScalarFloat m_beam_width, m_cutoff_angle, m_uv_factor;
+    ScalarFloat m_cos_beam_width, m_cos_cutoff_angle, m_inv_transition_width;
+};
+
+
+MTS_IMPLEMENT_CLASS_VARIANT(SpotLight, Emitter)
+MTS_EXPORT_PLUGIN(SpotLight, "Spot emitter")
+NAMESPACE_END(mitsuba)
--- a/src/emitters/tests/test_spot.py
+++ b/src/emitters/tests/test_spot.py
@@ -1,0 +1,178 @@
+import mitsuba
+import pytest
+import enoki as ek
+
+spectrum_strings = {
+    'd65':
+    """<spectrum version='2.0.0' name="intensity" type='d65'/>""",
+    'regular':
+    """<spectrum version='2.0.0' name="intensity" type='regular'>
+           <float name="lambda_min" value="500"/>
+           <float name="lambda_max" value="600"/>
+           <string name="values" value="1, 2"/>
+       </spectrum>"""
+}
+
+lookat_transforms = [
+    {
+        "pos": [0, 1, 0],
+        "target": [0, 0, 0],
+        "up": [1, 0, 0]
+    },
+    {
+        "pos": [0, 0, 0],
+        "target": [0, 0, 1],
+        "up": [0, 1, 0]
+    },
+    {
+        "pos": [0, 0, 1],
+        "target": [0, 0, 0],
+        "up": [0, -1, 0]
+    }
+]
+
+
+def create_emitter_and_spectrum(lookat, cutoff_angle, s_key='d65'):
+    from mitsuba.core.xml import load_string
+    emitter = load_string("""<emitter version='2.0.0' type='spot'>
+                                <float name='cutoff_angle' value='{ca}'/>
+                                <transform name='to_world'>
+                                    <lookat origin='{px}, {py}, {pz}' target='{tx}, {ty}, {tz}' up='{ux}, {uy}, {uz}'/>
+                                </transform>
+                                {s}
+                             </emitter>""".format(ca=cutoff_angle,
+                                                  px=lookat["pos"][0], py=lookat["pos"][1], pz=lookat["pos"][2],
+                                                  tx=lookat["target"][0], ty=lookat["target"][1], tz=lookat["target"][2],
+                                                  ux=lookat["up"][0], uy=lookat["up"][1], uz=lookat["up"][2],
+                                                  s=spectrum_strings[s_key]))
+    spectrum = load_string(spectrum_strings[s_key])
+    expanded = spectrum.expand()
+    if len(expanded) == 1:
+        spectrum = expanded[0]
+
+    return emitter, spectrum
+
+
+@pytest.mark.parametrize("spectrum_key", spectrum_strings.keys())
+@pytest.mark.parametrize("it_pos", [[2.0, 0.5, 0.0], [1.0, 0.5, -5.0], [0.0, 0.5, 0.0]])
+@pytest.mark.parametrize("wavelength_sample", [0.1, 0.3, 0.5, 0.7])
+@pytest.mark.parametrize("cutoff_angle", [20, 40, 60, 80])
+@pytest.mark.parametrize("lookat", lookat_transforms)
+def test_sample_direction(variant_scalar_spectral, spectrum_key, it_pos, wavelength_sample, cutoff_angle, lookat):
+    # Check the correctness of the sample_direction() method
+
+    import math
+    from mitsuba.core import sample_shifted, Transform4f
+    from mitsuba.render import SurfaceInteraction3f
+
+    cutoff_angle_rad = math.radians(cutoff_angle)
+    beam_width_rad = cutoff_angle_rad * 0.75
+    inv_transition_width = 1 / (cutoff_angle_rad - beam_width_rad)
+    emitter, spectrum = create_emitter_and_spectrum(
+        lookat, cutoff_angle, spectrum_key)
+    eval_t = 0.3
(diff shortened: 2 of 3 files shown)


src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+        return oss.str();
+    }
+
+    MTS_DECLARE_CLASS()
+private:
+    ref<Texture> m_intensity;
+	ref<Texture> m_texture;
+	Float m_beam_width, m_cutoff_angle, m_uv_factor;
+	Float m_cos_beam_width, m_cos_cutoff_angle, m_inv_transition_width;
+	Point3f m_position;
```
Speierers, line 170: Indentation typo

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+    MTS_IMPORT_BASE(Emitter, m_flags, m_medium, m_world_transform)
+    MTS_IMPORT_TYPES(Scene, Texture)
+
+    SpotLight(const Properties &props) : Base(props) {
+        m_intensity = props.texture<Texture>("intensity", Texture::D65(1.f));
+        m_cutoff_angle = props.float_("cutoff_angle", 20.0f);
+        m_beam_width = m_cutoff_angle * 3.0f/4.0f;
+
+        if (props.has_property("beam_width"))
+            m_beam_width = props.float_("beam_width");
```
Speierers, line 65:
```suggestion
        m_beam_width = props.float_("beam_width", m_cutoff_angle * 3.0f / 4.0f);
```

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+            << "]";
+        return oss.str();
+    }
+
+    MTS_DECLARE_CLASS()
+private:
+    ref<Texture> m_intensity;
+	ref<Texture> m_texture;
+	Float m_beam_width, m_cutoff_angle, m_uv_factor;
+	Float m_cos_beam_width, m_cos_cutoff_angle, m_inv_transition_width;
```
Speierers, line 169: Those members should be `ScalarFloat` instead of `Float`.

For variants other than `scalar_*`, `Float` represent a dynamic array of `float` values. Here a spot light object only represent a single spot light so those members should be scalar.

src/emitters/spot.cpp
Speierers, line 170: Similarly here, it should be `ScalarPoint3f`

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+   - Specifies the maximum radiant intensity at the center in units of power per unit steradian. (Default: 1).
+ * - cutoff_angle
+   - |float|
+   - Cutoff angle, beyond which the spot light is completely black (Default: 20 degrees)
+ * - beam_width
+   - |float|
+   - Subtended angle of the central beam portion (Default: :math:`cutoff_angle \times 3/4`)
+ * - texture
+   - |texture|
+   - An optional texture to be projected along the spot light
```
Speierers, line 29: In Mitsuba 2 spectrum properties (like `intensity`) are can always be spatially-variant. Therefore this `texture` is redondant. Maybe we could remove it? Or would it break backward compatibility?

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+        SurfaceInteraction3f si = zero<SurfaceInteraction3f>();
+        si.wavelengths = wavelengths;
+        Spectrum result = m_intensity->eval(si, active);
+
+        const Transform4f &trafo = m_world_transform->eval(time, active);
+        Vector3f local_dir = trafo.inverse() * -d;
+        local_dir /= norm(local_dir);
+        const Float cos_theta = local_dir.z();
+
+        si.uv = Point2f(0.5f + 0.5f * local_dir.x() / (local_dir.z() * m_uv_factor), 0.5f + 0.5f * local_dir.y() / (local_dir.z() * m_uv_factor));
```
Speierers, line 88: Format this on multiple lines. See `.clang-format`

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+        result *= m_texture->eval(si, active);
+        
+        return select(cos_theta <= m_cos_cutoff_angle, Spectrum(0.0f), select(cos_theta >= m_cos_beam_width, result, result * ((m_cutoff_angle - acos(cos_theta)) * m_inv_transition_width)));
```
Speierers, line 91: Break this on multiple lines for clarity 😉

src/emitters/spot.cpp
Speierers: This being said, if the plan is to be able to optimize those parameters using autodiff, then they should indeed be `Float`.

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+                        unpolarized<Spectrum>(spec_weight) * (2 * math::Pi<Float> * (1 - (sin(m_cutoff_angle) - sin(m_beam_width))/(m_cutoff_angle - m_beam_width)))) };
+    }
+
+    std::pair<DirectionSample3f, Spectrum> sample_direction(const Interaction3f &it,
+                                                            const Point2f & /*sample*/,
+                                                            Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleDirection, active);
+
+        auto trafo = m_world_transform->eval(it.time, active);
+
```
Speierers, line 119: Here we should sample a direction on a cone and evaluate the corresponding pdf.

Please have a look at the Mitsuba 1 implementation, it should be very similar to that.

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+        ds.dist  = norm(ds.d);
+        Float inv_dist = rcp(ds.dist);
+        ds.d *= inv_dist;
+
+        Spectrum spec = falloff_curve(ds.d, it.time, it.wavelengths, active) * (inv_dist * inv_dist);
+        return { ds, unpolarized<Spectrum>(spec) };
+    }
+
+    Float pdf_direction(const Interaction3f &, const DirectionSample3f &, Mask) const override {
+        return 0.f;
```
Speierers, line 138: I guess here it should return warp::squareToUniformConePdf(m_cosCutoffAngle) instead?

src/emitters/spot.cpp
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    Spectrum eval(const SurfaceInteraction3f &, Mask) const override { return 0.f; }
```
Speierers, line 141: Should evaluated falloff here no?

src/emitters/spot.cpp
mehrab2603: Should I change line 125 to
`ds.pdf   = warp::square_to_uniform_cone_pdf(warp::square_to_uniform_cone(sample, m_cos_cutoff_angle), m_cos_cutoff_angle);`

and then at line 138, return ds.pdf?

src/emitters/spot.cpp
mehrab2603: I think we should let these be optimizable. Is there any reason we should not?

src/emitters/spot.cpp
mehrab2603: I tried this initially, but I was getting the following error. Not sure how to fix it.

`../include/mitsuba/core/properties.h:193:18: note: candidate function not viable: no known conversion from 'enoki::DiffArray<CUDAArray<float> >' to 'const mitsuba::Properties::Float' (aka 'const float') for 2nd argument const Float& float_(const std::string &name, const Float &def_val) const;`

src/emitters/spot.cpp
mehrab2603, line 91: How about this?

```suggestion
        return select(cos_theta <= m_cos_cutoff_angle, Spectrum(0.0f),
                select(cos_theta >= m_cos_beam_width, result,
                result * ((m_cutoff_angle - acos(cos_theta)) * m_inv_transition_width)));
```

src/emitters/spot.cpp
mehrab2603, line 88: How about this? clang-format is not working for me for some reason.

```suggestion
        si.uv = Point2f(0.5f + 0.5f * local_dir.x() / (local_dir.z() * m_uv_factor),
                        0.5f + 0.5f * local_dir.y() / (local_dir.z() * m_uv_factor));
```

src/emitters/spot.cpp
mehrab2603: This was there in Mitsuba 1 and so it would break backward compatibility indeed. Should it be removed?

src/emitters/spot.cpp
Speierers: indeed this is related to the comments about the members

src/emitters/spot.cpp
Speierers: For now let's have them as `ScalarFloat` for simplicity. Other plugins also have tons of properties that are not optimizable ATM, we will do a pass over all of those in the future.

src/emitters/spot.cpp
Speierers: You will need to assign the result of `square_to_uniform_cone(sample, m_cos_cutoff_angle)` to `ds.d` as well.

And `pdf_direction()` should recompute the PDF given `ds.d` as it might be sampled from another distribution.

src/emitters/spot.cpp
Speierers: Better

src/emitters/spot.cpp
Speierers: This is still quite obscur. Could we store intermediate results in other variables?

src/emitters/spot.cpp
mehrab2603: Converting `m_cos_cutoff_angle` to `ScalarFloat` causes `square_to_uniform_cone` to complain. Should I leave just `m_cos_cutoff_angle` to `Float`?

src/emitters/spot.cpp
mehrab2603: But I am not given any `sample` in `pdf_direction()`. How do I call get a sample for  calling `square_to_uniform_cone_pdf()`?

src/emitters/spot.cpp
Speierers: No, you can simply cast it as a ` Float` when calling the function.

src/emitters/spot.cpp
Speierers: For now let's only keep `m_intensity` internally. In the constructor, if only `texture` if provided (check with `has_property()`) then you use it to set `m_intensity`. If both are provided, it should throw and error for now. 

In the future, we could for instance multiply those two textures, but it isn't possible to do for now.

src/emitters/spot.cpp
Speierers: `square_to_uniform_cone_pdf` takes a 3D vector (the direction) not a 2D point on the unit square. You have this direction in `ds.d`. Also, you might have to transform it using the `m_world_transform` matrix.

mehrab2603:
I have made your suggested changes except for removing m_texture and just using m_intensity. Having m_intensity separately allows the user to adjust the brightness of the light while m_texture can be used to project an image. Is this achievable in the scene XML with just the single bitmap type texture tag? Still not sure about a lot of things. Please let me know how things look at the moment.

src/emitters/spot.cpp
```diff
@@ -0,0 +1,188 @@
...
+        m_beam_width = props.float_("beam_width", m_cutoff_angle * 3.0f / 4.0f);
+        m_cutoff_angle = deg_to_rad(m_cutoff_angle);
+        m_beam_width = deg_to_rad(m_beam_width);
+        m_inv_transition_width = 1.0f / (m_cutoff_angle - m_beam_width);
+        m_cos_cutoff_angle = cos(m_cutoff_angle);
+        m_cos_beam_width = cos(m_beam_width);
+        Assert(m_cutoff_angle >= m_beam_width);
+        m_uv_factor = tan(m_cutoff_angle);
+        m_texture = props.texture<Texture>("texture", Texture::D65(1.f));
+        m_position = ScalarPoint3f(0.0f);
```
Speierers, line 71: Do we really need this?

src/emitters/spot.cpp
```diff
@@ -0,0 +1,188 @@
...
+        Vector3f local_dir = -d;
+        local_dir /= norm(local_dir);
+        const Float cos_theta = local_dir.z();
+
+        si.uv = Point2f(0.5f + 0.5f * local_dir.x() / (local_dir.z() * m_uv_factor),
+                        0.5f + 0.5f * local_dir.y() / (local_dir.z() * m_uv_factor));
+        result *= m_texture->eval(si, active);
+
+        auto beam_res = select(cos_theta >= m_cos_beam_width, result,
+                        result * ((m_cutoff_angle - acos(cos_theta)) * m_inv_transition_width));
```
Speierers, line 89: Indentation / style. Please try to use clang-format for those long lines.

src/emitters/spot.cpp
```diff
@@ -0,0 +1,188 @@
...
+        auto trafo = m_world_transform->eval(it.time, active);
+
+        DirectionSample3f ds;
+        ds.p        = trafo * m_position;
+        ds.n        = 0.f;
+        ds.uv       = 0.f;
+        ds.time     = it.time;
+        ds.delta    = true;
+        ds.object   = this;
+        ds.d        = ds.p - it.p;
```
Speierers, line 128: I guess we should sample a direction in the cone here no?

src/emitters/spot.cpp
```diff
@@ -0,0 +1,188 @@
...
+    std::pair<DirectionSample3f, Spectrum> sample_direction(const Interaction3f &it,
+                                                            const Point2f &/*sample*/,
+                                                            Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleDirection, active);
+
+        auto trafo = m_world_transform->eval(it.time, active);
+
+        DirectionSample3f ds;
+        ds.p        = trafo * m_position;
+        ds.n        = 0.f;
```
Speierers, line 123: This could be the main direction of the cone

src/emitters/spot.cpp
Speierers: I take it back, I got all confused with `sample_ray`. Sorry.

src/emitters/spot.cpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    Spectrum eval(const SurfaceInteraction3f &si, Mask active) const override { 
+        auto trafo = m_world_transform->eval(si.time, active);
+        Vector3f d = trafo * m_position - si.p;
+        Float inv_dist = rcp(norm(d));
+        d *= inv_dist;
+        Spectrum falloff_spec = falloff_curve(trafo.inverse() * d, si.wavelengths, active);
+
+        return unpolarized<Spectrum>(falloff_spec * inv_dist * inv_dist);
```
Speierers, line 152: I just realized I was wrong. Those two methods should indeed return `0.f`. I kept thinking about this implementation this other way (e.g. `sample_ray`). This should make things easier now ;)

src/emitters/spot.cpp
mehrab2603: No we don't. I'll remove it.

src/emitters/spot.cpp
mehrab2603: Couldn't get clang-format to work on VSCode. Sorry.

src/emitters/spot.cpp
mehrab2603: Should the `ds.pdf` at line 133 be set to 1 like the implementations of point and directional emitters?

src/emitters/spot.cpp
mehrab2603: Okay, I will make `eval()` and `pdf_direction()` make return 0. But just out of curiosity - why do you think these should return 0 and why did you think they should not earlier? Just want to learn as I am making another kind of light and knowing this will help.

src/emitters/spot.cpp
Speierers: Yes

src/emitters/spot.cpp
Speierers:
```suggestion
        auto beam_res = select(cos_theta >= m_cos_beam_width, result,
                               result * ((m_cutoff_angle - acos(cos_theta)) * m_inv_transition_width));
```

src/emitters/spot.cpp
Speierers: First I was thinking of `pdf_direction()` as the probability of sampling a specific direction **from the light source**. (e.g. what `sample_ray()` does).

But this method actually compute the probability of sampling a direction **towards the light source** from a surface (e.g. what `sample_direction()` does).

As this light source is a `DeltaPositionEmitter` (just a point), it is impossible to hit it *by chance*. In other words, the probability of sampling a direction towards this emitter via other methods than calling `sample_direction()` on it is `0.f`.

Does this make more sense to you?

mehrab2603: Yes, crystal clear. Thanks a lot.

mehrab2603:
Made changes as suggested.

src/emitters/spot.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    std::pair<DirectionSample3f, Spectrum> sample_direction(const Interaction3f &it,
+                                                            const Point2f &/*sample*/,
+                                                            Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::EndpointSampleDirection, active);
+
+        auto trafo = m_world_transform->eval(it.time, active);
+
+        DirectionSample3f ds;
+        ds.p        = trafo * ScalarPoint3f(0.0f);
```
Speierers, line 123: You could use `trafo.translation()` here

src/emitters/spot.cpp
```diff
@@ -0,0 +1,179 @@
...
+        m_flags = EmitterFlags::DeltaPosition | EmitterFlags::SpatiallyVarying;
+    }
+
+    inline Spectrum falloff_curve(const Vector3f &d, Wavelength wavelengths, Mask active) const {
+        SurfaceInteraction3f si = zero<SurfaceInteraction3f>();
+        si.wavelengths = wavelengths;
+        Spectrum result = m_intensity->eval(si, active);
+
+        Vector3f local_dir = -d;
+        local_dir /= norm(local_dir);
```
Speierers, line 80:
```suggestion
        Vector3f local_dir = -normalized(d);
```

src/emitters/spot.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+        si.uv = Point2f(0.5f + 0.5f * local_dir.x() / (local_dir.z() * m_uv_factor),
+                        0.5f + 0.5f * local_dir.y() / (local_dir.z() * m_uv_factor));
+        result *= m_texture->eval(si, active);
+
+        auto beam_res = select(cos_theta >= m_cos_beam_width, result,
+                               result * ((m_cutoff_angle - acos(cos_theta)) * m_inv_transition_width));
+        auto beam_and_cutoff_res = select(cos_theta <= m_cos_cutoff_angle, Spectrum(0.0f), beam_res);
+
+        return beam_and_cutoff_res;
```
Speierers, line 91:
```suggestion
        return select(cos_theta <= m_cos_cutoff_angle, Spectrum(0.0f), beam_res);
```

src/emitters/spot.cpp
```diff
@@ -0,0 +1,179 @@
...
+        Vector3f local_dir = warp::square_to_uniform_cone(spatial_sample, (Float)m_cos_cutoff_angle);
+        Float pdf_dir = warp::square_to_uniform_cone_pdf(local_dir, (Float)m_cos_cutoff_angle);
+
+        // 2. Sample spectrum
+        auto [wavelengths, spec_weight] = m_intensity->sample(
+            zero<SurfaceInteraction3f>(),
+            math::sample_shifted<Wavelength>(wavelength_sample), active);
+        auto falloff_spec = falloff_curve(local_dir, wavelengths, active);
+
+        return { Ray3f(trafo * ScalarPoint3f(0.0f), trafo * local_dir, time, wavelengths),
```
Speierers, line 111: Use `trafo.translation()`

src/emitters/spot.cpp
```diff
@@ -0,0 +1,179 @@
...
+        m_cutoff_angle = props.float_("cutoff_angle", 20.0f);
+        m_beam_width = props.float_("beam_width", m_cutoff_angle * 3.0f / 4.0f);
+        m_cutoff_angle = deg_to_rad(m_cutoff_angle);
+        m_beam_width = deg_to_rad(m_beam_width);
+        m_inv_transition_width = 1.0f / (m_cutoff_angle - m_beam_width);
+        m_cos_cutoff_angle = cos(m_cutoff_angle);
+        m_cos_beam_width = cos(m_beam_width);
+        Assert(m_cutoff_angle >= m_beam_width);
+        m_uv_factor = tan(m_cutoff_angle);
+        m_texture = props.texture<Texture>("texture", Texture::D65(1.f));
```
Speierers, line 70: Let's ensure `intensity` isn't a spatially varying texture and `texture` is when both provided. For this you will need the latest commit on master that adds the `Texture::is_spatially_varying()` method.

src/emitters/tests/test_spot.py
```diff
@@ -0,0 +1,135 @@
...
+    # Check the correctness of the sample_ray() method
+
+    import math
+    from mitsuba.core import Vector3f, warp, sample_shifted, Transform4f
+    from mitsuba.render import SurfaceInteraction3f
+
+    emitter_pos = [0, 1, 0]
+    emitter_target = [0, 0, 0]
+    emitter_up = [1, 0, 0]
+    cutoff_angle_deg = 40
```
Speierers, line 50: Could parameterize the test on those as well?

src/emitters/tests/test_spot.py
```diff
@@ -0,0 +1,135 @@
...
+    cos_cutoff_angle_rad = math.cos(math.radians(cutoff_angle_deg))
+    emitter, spectrum = create_emitter_and_spectrum(
+        emitter_pos, emitter_target, emitter_up, cutoff_angle_deg, spectrum_key)
+
+    time = 0.5
+    wavelength_sample = [0.5, 0.33, 0.1]
+    pos_sample = [[0.4, 0.5, 0.3], [0.1, 0.4, 0.9]]
+    dir_sample = pos_sample  # not being used anyway
```
Speierers, line 58: Maybe iterate here a couple times over different samples (could use `np.random.rand()` with a fixed seed)

src/emitters/tests/test_spot.py
```diff
@@ -0,0 +1,135 @@
...
+@pytest.mark.parametrize("spectrum_key", spectrum_strings.keys())
+def test_sample_direction(variant_scalar_spectral, spectrum_key):
+    # Check the correctness of the sample_direction() method
+
+    from mitsuba.render import SurfaceInteraction3f
+
+    emitter_pos = [0, 1, 0]
+    emitter_target = [0, 0, 0]
+    emitter_up = [1, 0, 0]
+    cutoff_angle = 40
```
Speierers, line 88: Parameterize this

src/emitters/tests/test_spot.py
```diff
@@ -0,0 +1,135 @@
...
+    emitter, spectrum = create_emitter_and_spectrum(
+        emitter_pos, emitter_target, emitter_up, cutoff_angle, spectrum_key)
+
+    # Check that eval() return the same values as the 'radiance' spectrum
+    it = SurfaceInteraction3f.zero(3)
+    assert ek.allclose(emitter.eval(it), spectrum.eval(it))
```
Speierers, line 130: Does `spectrum.eval(it)` always return `0.f`?

src/emitters/tests/test_spot.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Check that incident direction in the illuminated direction is zero (because hitting a delta light is impossible)
+    it = SurfaceInteraction3f.zero(3)
+    it.wi = [0, 1, 0]
+    assert ek.allclose(emitter.eval(it), 0.)
```
Speierers, line 135: Could test `.pdf(it)==0.0` at the same time

src/emitters/spot.cpp
```diff
@@ -57,7 +57,23 @@ class SpotLight final : public Emitter<Float, Spectrum> {
     MTS_IMPORT_TYPES(Scene, Texture)
 
     SpotLight(const Properties &props) : Base(props) {
-        m_intensity = props.texture<Texture>("intensity", Texture::D65(1.f));
+        if (props.has_property("intensity") && props.has_property("texture")) {
+            m_intensity = props.texture<Texture>("intensity");
+            m_texture = props.texture<Texture>("texture");
+            
+            if (m_intensity->is_spatially_varying())
```
Speierers, line 67: Actually, I think if we keep the two parameters, then `intensity` should never be spatially varying, even when `texture` isn't specified.

mehrab2603: Should texture be always varying even if intensity is not specified?

Speierers: Yes I would say so

Speierers:
Would it be possible to also add a test scene with a spot light in the `resources/data/tests/scene` folder? For this you will need to open a PR on the `data` repo, which I will merge first, and then we will be able to update the submodule in this one.

Looking forward to merging this PR! 🚀

mehrab2603:
Done.

src/emitters/tests/test_spot.py
```diff
@@ -37,97 +36,171 @@ def create_emitter_and_spectrum(pos, target, up, cutoff_angle, s_key='d65'):
...
+        "pos": [0, 0, 0],
+        "target": [0, 0, 1],
+        "up": [0, 1, 0]
+    },
+    {
+        "pos": [0, 0, 1],
+        "target": [0, 0, 0],
+        "up": [0, -1, 0]
+    }
+])
```
Speierers, line 131: Could we avoid duplicate this 3 times? 😉

Speierers:
Great work! I will merge the other PR for the rendering test.

Could you please address the "minor" comment I added regarding code duplication in the unit test, and then squash all those commits into a single commit?
You will also be able to update the `resources/data` submodule once the other PR is merged.

Then we are ready to launch 🚀

Speierers:
One last request: could we add maybe 2 renders in the documentation to illustrate the fonctionallity of the spot light? E.g. one without texture and one with. Or just varying the cutoff angle?

For this you will need to add stuff in the `resources/data/docs/scenes` and point to the images in the docstring at the top of the plugin.

This would be great!

Also you can update the submodule now!

mehrab2603:
Should I open another PR in the resources/data to add those files?

Speierers:
I just merge the PR on `resources/data`, you can update the data submodule

mehrab2603:
Think I updated the submodule correctly and squashed all the commits.

Speierers:
Thanks @mehrab2603 for your hard work. I will merge this 🚀

mehrab2603:
Thank you @Speierers for guiding me through this process with patience. I learned a lot. Would you be interested in adding a projector like emitter with parameters similar to a perspective camera (fov, fov_axis, focal_length)? Outputs look like this.

output

Speierers:
Sure, if you open another PR I will be happy to help you on this as well 😉

PR #102, opened 2020-04-22
minor fix for dielectric plugin documentation / homogeneous medium parametrization

dnageo:
I changed the parameter names as well as the values as the following range is now expected in xml:
```
[xml.cpp:216]   [SRGBReflectanceSpectrum] Invalid RGB reflectance value [4, 4, 2], must be in the range [0, 1]!.
```
Please check if those values still describe a *slightly absorbing piece of glass* as -- as far as i understood -- _sigma_s_ is now computed from _sigma_t_ and _albedo_ in *src/media/homogeneous.cpp*.

### DIFF
--- a/src/bsdfs/dielectric.cpp
+++ b/src/bsdfs/dielectric.cpp
@@ -80,8 +80,8 @@ An example of how one might describe a slightly absorbing piece of glass is show
         </bsdf>
 
         <medium type="homogeneous" name="interior">
-            <rgb name="sigma_s" value="0, 0, 0"/>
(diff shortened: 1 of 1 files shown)


dvicini:
Hi, 
Thanks for pull request. I think the values you have are slightly wrong. To match the values from before the change you need
```
	<float name="density" value="4"/>
	<rgb name="sigma_t" value="1, 1, 0.5"/>
	<rgb name="albedo" value="0.0, 0.0, 0.0"/>
```
where density scales the sigmat values. Please use these values instead

PR #106, opened 2020-04-23
Minor fix for roughplastic plugin

neverfelly:
I compare it with Mitsuba1 and think this might be a small typo.


Speierers:
Your are correct, thanks for reporting this! I will merge it.

PR #117, opened 2020-05-01
Projector emitter

mehrab2603:
This is an initial implementation of a projector like emitter. `sample_ray` is not implemented because I am not sure how handle sampling from a rectangle. Tests will be added later.

output


Speierers:
Hi @mehrab2603 ,

Thanks for this PR.

Could you explain the fundamental difference between this plugin and the `spot` light plugin?

They look very similar to me, I wonder whether we couldn't somehow extend the existing `spot` plugin so it fulfill the purpose of this plugin as well.

mehrab2603:
This basically simulates a projector which is basically an inverse perspective camera and so you control FOV and focal length. You don't have those in spot light. Of course, spot can be extended to add those, but spot should, in principle, not worry about those parameters.

Speierers:
@mehrab2603, indeed this should be treated as a different plugin.

Let's make sure this emitter mimic exactly the behavior of the perspective camera then. A lot of the `perspective.cpp` code could be shared then. I think we should move this to a seperate header. If I remember corretly there is even a `TODO` somewhere as this code is already duplicated in `thinlens.cpp`. We should include this refacting in this PR as well if you don't mind?

wjakob:
Dear @mehrab2603,

thank you for your pull request. I looked into this today and decided to follow a slightly different implementation that is closer to the perspective camera interface (since those two are really reciprocal versions of each other). This is in 'master' as of now.

Thanks,
Wenzel

PR #121, opened 2020-05-04
Fix missing ray update in radiancemeter plugin

leroyvn:
This small PR fixes a bug in the `radiancemeter` plugin where an omitted ray update would lead to invalid intersections.


Speierers:
Thanks @leroyvn , I will merge this right away 🚀

merlinND:
Good catch, this is a classic bug :)

PR #123, opened 2020-05-05
Implicit shape refactor

4str0m:
Unify implicit shapes implementation.

### DIFF
--- a/src/shapes/cylinder.cpp
+++ b/src/shapes/cylinder.cpp
@@ -4,11 +4,8 @@
 #include <mitsuba/core/transform.h>
 #include <mitsuba/core/util.h>
 #include <mitsuba/core/warp.h>
-#include <mitsuba/render/bsdf.h>
-#include <mitsuba/render/emitter.h>
 #include <mitsuba/render/fwd.h>
 #include <mitsuba/render/interaction.h>
-#include <mitsuba/render/sensor.h>
 #include <mitsuba/render/shape.h>
 
 NAMESPACE_BEGIN(mitsuba)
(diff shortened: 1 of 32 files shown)


src/shapes/rectangle.cpp
```diff
@@ -107,7 +109,9 @@ class Rectangle final : public Shape<Float, Spectrum> {
     }
 
     ScalarFloat surface_area() const override {
-        return m_du * m_dv;
+        // First compute height of the parallelogram
+        ScalarFloat h = sqrt(sqr(m_dv) - sqr(dot(m_dv * m_frame.t, m_frame.s)));
```
Speierers, line 113: Should use `cross`

src/shapes/cylinder.cpp
```diff
@@ -217,7 +226,7 @@ class Cylinder final : public Shape<Float, Spectrum> {
 
         PositionSample3f ps;
         ps.p     = m_to_world.transform_affine(p);
-        ps.n     = normalize(n);
+        ps.n     = normalize(m_to_world * n);
```
Speierers, line 229: `transform_affine()`?

src/shapes/cylinder.cpp
```diff
@@ -206,9 +215,9 @@ class Cylinder final : public Shape<Float, Spectrum> {
 
         auto [sin_theta, cos_theta] = sincos(2.f * math::Pi<Float> * sample.y());
 
-        Point3f p(cos_theta * m_radius,
-                 sin_theta * m_radius,
-                 sample.x() * m_length);
+        Point3f p(cos_theta,
+                 sin_theta,
```
Speierers, line 219: indentation

src/shapes/cylinder.cpp
```diff
@@ -82,118 +82,127 @@ class Cylinder final : public Shape<Float, Spectrum> {
     using typename Base::ScalarSize;
 
     Cylinder(const Properties &props) : Base(props) {
-        m_radius = props.float_("radius", 1.f);
+        bool has_to_world = props.has_property("to_world");
 
-        ScalarPoint3f p0 = props.point3f("p0", ScalarPoint3f(0.f, 0.f, 0.f)),
-                      p1 = props.point3f("p1", ScalarPoint3f(0.f, 0.f, 1.f));
+        if (!has_to_world) {
```
Speierers, line 87: Could we have both the `to_world` transform and the `p0`, `p1` properties at the same time?
In which case we should first build a pre_transform with the `p0` `p1` properties and later multiply the `to_world` transform. See how it is done in `sphere.cpp`.

PR #130, opened 2020-05-07
Minor changes to directional

leroyvn:
Here are a couple of changes I made to `directional` I'd like some feedback on:

1. Since only 2 samples are used by `sample_ray()`, I set `m_needs_sample_3` to `false`.
2. When writing this plugin, I copied the importance weight value from `constant` and set its factor to . However, it seems like it is proportional to the bounding sphere's surface area  (this is in the tweaks I made). I'm not quite sure about this, I actually don't understand how this value is chosen so I'd like to know if it actually matters :)

I also happened to test a distant directional sensor plugin with more samples so I added these to `directional`'s tests. Which makes me think: would such a sensor be of use to the community?

### DIFF
--- a/src/emitters/tests/test_directional.py
+++ b/src/emitters/tests/test_directional.py
@@ -130,25 +130,37 @@ def test_sample_direction(variant_scalar_spectral, spectrum_key, direction):
     assert ek.allclose(res, spec)
 
 
-@pytest.mark.parametrize("spatial_sample", [[0.85, 0.13], [0.16, 0.50], [0.00, 1.00]])
 @pytest.mark.parametrize("direction", [[0, 0, -1], [1, 1, 1], [0, 0, 1]])
-def test_sample_ray(variant_scalar_rgb, spatial_sample, direction):
+def test_sample_ray(variant_scalar_rgb, direction):
     import enoki as ek
     from mitsuba.core import Vector2f, Vector3f
 
     emitter = make_emitter(direction=direction)
     direction = Vector3f(direction)
-    
+
     time = 1.0
     wavelength_sample = 0.3
     directional_sample = [0.3, 0.2]
 
-    ray, wavelength = emitter.sample_ray(
-        time, wavelength_sample, spatial_sample, directional_sample)
+    for spatial_sample in [
+        [0.85, 0.13],
+        [0.16, 0.50],
+        [0.00, 1.00],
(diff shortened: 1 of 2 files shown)


src/emitters/tests/test_directional.py
```diff
@@ -130,7 +130,19 @@ def test_sample_direction(variant_scalar_spectral, spectrum_key, direction):
...
+    [0.85, 0.13], 
+    [0.16, 0.50], 
+    [0.00, 1.00],
+    [0.32, 0.87], 
+    [0.16, 0.44],
+    [0.17, 0.44], 
+    [0.22, 0.81],
+    [0.12, 0.82], 
+    [0.99, 0.42],
+    [0.72, 0.40],
```
Speierers, line 143: Would you mind moving this array of samples within the test so if it fails, we don't get 30 tests failing on the same bug?

Speierers:
Hi @leroyvn , thanks for this PR!

> Which makes me think: would such a sensor be of use to the community?
I am not sure to understand what kind of sensor you are talking about here?

> When writing this plugin, I copied the importance weight value from constant...
Indeed the `sample_ray` should return the power which is inversly proportional to the surface area of the emitter. Would this mean that there is a typo in `constant`?

leroyvn:
> I am not sure to understand what kind of sensor you are talking about here?

What I have in mind is a sensor "at infinite distance" which records the radiance leaving the scene in a given direction. We use that to compute the BRDF of a complex scene when simulating satellite observations.

> Indeed the `sample_ray` should return the power which is inversly proportional to the surface area of the emitter. Would this mean that there is a typo in `constant`?

I didn't know it was the power, thanks for clarifying that to me. It's clear now: `constant` returns  (_r_ being the bounding sphere's radius), which is the total emitted spectral power (in W/nm if we use that unit convention). No typo, I was mistaken!

So I assume `directional` should actually return , right?

src/emitters/tests/test_directional.py
leroyvn: Fixed!

Speierers:
> So I assume directional should actually return
Correct!

Could we squash those 3 commits into one, then I will merge it 🚀 

Regarding the infinite sensor, please go ahead an open another PR if you already have an implementation ;)

leroyvn:
Here we go, thanks for the feedback and advice!

PR #131, opened 2020-05-11
Write python dictionary to XML

bathal1:
Implementation of the XML writing routine, which can be used with python dictionaries created either by users or by the blender exporter addon

### DIFF
--- a/src/libcore/tests/test_write_xml.py
+++ b/src/libcore/tests/test_write_xml.py
@@ -1,0 +1,458 @@
+import enoki as ek
+import pytest
+import mitsuba
+import os
+from shutil import rmtree
+
+from mitsuba.python.test.util import fresolver_append_path
+
+@fresolver_append_path
+def test01_xml_save_plugin(variant_scalar_rgb):
+    from mitsuba.core import xml
+    from mitsuba.core import Thread
+    from mitsuba.python.xml import dict_to_xml
+    fr = Thread.thread().file_resolver()
+    # Add the path to the mitsuba root folder, so that files are always saved in mitsuba/resources/...
+    # This way we know where to look for the file in case the unit test fails
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict01/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    scene_dict = {
+            "type": "sphere",
+            "center" : [0, 0, -10],
+            "radius" : 10.0,
+        }
+    dict_to_xml(scene_dict, filepath)
+    s1 = xml.load_dict({
+        'type': 'scene',
+        'sphere':{
+            "type": "sphere",
+            "center" : [0, 0, -10],
+            "radius" : 10.0,
+            }
+        })
+    s2 = xml.load_file(filepath)
+    assert str(s1) == str(s2)
+    rmtree(os.path.split(filepath)[0])
+
+@fresolver_append_path
+def test02_xml_missing_type(variant_scalar_rgb):
+    from mitsuba.python.xml import dict_to_xml
+    from mitsuba.core import Thread
+    fr = Thread.thread().file_resolver()
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict02/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    scene_dict = {
+        'my_bsdf':{
+            'type':'diffuse'
+        }
+    }
+    with pytest.raises(ValueError) as e:
+        dict_to_xml(scene_dict, filepath)
+    e.match("Missing key: 'type'!")
+    rmtree(os.path.split(filepath)[0])
+
+@fresolver_append_path
+def test03_xml_references(variant_scalar_rgb):
+    from mitsuba.core import xml
+    from mitsuba.python.xml import dict_to_xml
+    from mitsuba.core import Thread
+    fr = Thread.thread().file_resolver()
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict03/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    scene_dict = {
+        'type': 'scene',
+        'bsdf1': {
+            'type': 'diffuse',
+            'reflectance': {
+                'type': 'rgb',
+                'value': [1.0, 0.0, 0.0]
+            }
+        },
+        'texture1': {
+            'type': 'checkerboard'
+        },
+        'bsdf4': {
+            'type': 'dielectric',
+            'specular_reflectance': {
+                'type': 'ref',
+                'id': 'texture1'
+            },
+            'id': 'bsdf2'
+        },
+        'shape1': {
+            'type': 'sphere',
+            'bsdf': {
+                'type': 'ref',
+                'id': 'bsdf1'
+            }
+        },
+        'shape2': {
+            'type': 'cylinder',
+            'bsdf': {
+                'type': 'ref',
+                'id': 'bsdf2'
+            }
+        }
+    }
+
+    s1 = xml.load_dict(scene_dict)
+    dict_to_xml(scene_dict, filepath)
+    s2 = xml.load_file(filepath)
+
+    assert str(s1.shapes()[0].bsdf()) == str(s2.shapes()[0].bsdf())
+    rmtree(os.path.split(filepath)[0])
+
+@fresolver_append_path
+def test04_xml_point(variant_scalar_rgb):
+    from mitsuba.core import xml, Point3f
+    from mitsuba.python.xml import dict_to_xml
+    from mitsuba.core import Thread
+    import numpy as np
+    fr = Thread.thread().file_resolver()
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict04/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    scene_dict = {
+        'type': 'scene',
+        'light':{
+            'type': 'point',
+            'position': [0.0, 1.0, 2.0]
+        }
+    }
+    dict_to_xml(scene_dict, filepath)
+    s1 = xml.load_file(filepath)
+    scene_dict = {
+        'type': 'scene',
+        'light':{
+            'type': 'point',
+            'position': Point3f(0, 1, 2)
+        }
+    }
+    dict_to_xml(scene_dict, filepath)
+    s2 = xml.load_file(filepath)
+    scene_dict = {
+        'type': 'scene',
+        'light':{
+            'type': 'point',
+            'position': np.array([0, 1, 2])
+        }
+    }
+    dict_to_xml(scene_dict, filepath)
+    s3 = xml.load_file(filepath)
+
+    assert str(s1) == str(s2)
+    assert str(s1) == str(s3)
+    rmtree(os.path.split(filepath)[0])
+
+@fresolver_append_path
+def test05_xml_split(variant_scalar_rgb):
+    from mitsuba.core import xml, Point3f
+    from mitsuba.python.xml import dict_to_xml
+    from mitsuba.core import Thread
+    import numpy as np
+    fr = Thread.thread().file_resolver()
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict05/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    scene_dict = {
+        'type': 'scene',
+        'bsdf1': {
+            'type': 'diffuse'
+        },
+        'envmap': {
+            'type': 'envmap',
+            'filename': 'resources/data/envmap/museum.exr'
+        },
+        'shape':{
+            'type': 'sphere',
+            'bsdf': {
+                'type': 'ref',
+                'id': 'bsdf1'
+            }
+        }
+    }
+    dict_to_xml(scene_dict, filepath)
+    s1 = xml.load_file(filepath)
+    dict_to_xml(scene_dict, filepath, split_files=True)
+    s2 = xml.load_file(filepath)
+    assert str(s1) == str(s2)
+    rmtree(os.path.split(filepath)[0])
+
+@fresolver_append_path
+def test06_xml_duplicate_id(variant_scalar_rgb):
+    from mitsuba.python.xml import dict_to_xml
+    from mitsuba.core import Thread
+    fr = Thread.thread().file_resolver()
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict06/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    scene_dict = {
+        'type': 'scene',
+        'my-bsdf': {
+            'type': 'diffuse'
+        },
+        'bsdf1': {
+            'type': 'roughdielectric',
+            'id': 'my-bsdf'
+        }
+    }
+    with pytest.raises(ValueError) as e:
+        dict_to_xml(scene_dict, filepath)
+    e.match("Id: my-bsdf is already used!")
+    rmtree(os.path.split(filepath)[0])
+
+@fresolver_append_path
+def test07_xml_invalid_ref(variant_scalar_rgb):
+    from mitsuba.python.xml import dict_to_xml
+    from mitsuba.core import Thread
+    fr = Thread.thread().file_resolver()
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict07/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    scene_dict = {
+        'type': 'scene',
+        'bsdf': {
+            'type': 'diffuse'
+        },
+        'sphere': {
+            'type': 'sphere',
+            'bsdf': {
+                'type': 'ref',
+                'id': 'my-bsdf'
+            }
+        }
+    }
+    with pytest.raises(ValueError) as e:
+        dict_to_xml(scene_dict, filepath)
+    e.match("Id: my-bsdf referenced before export.")
+    rmtree(os.path.split(filepath)[0])
+
+@fresolver_append_path
+def test08_xml_defaults(variant_scalar_rgb):
+    from mitsuba.python.xml import dict_to_xml
+    from mitsuba.core import Thread
+    from mitsuba.core.xml import load_dict, load_file
+    fr = Thread.thread().file_resolver()
+    mts_root = str(fr[len(fr)-1])
+    filepath = os.path.join(mts_root, 'resources/data/scenes/dict08/dict.xml')
+    fr.append(os.path.dirname(filepath))
+    spp = 250
+    resx = 1920
+    resy = 1080
+    scene_dict = {
+        'type': 'scene',
+        'cam': {
+            'type': 'perspective',
+            'fov_axis': 'x',
+            'fov': 35,
+            'sampler': {
+                'type': 'independent',
+                'sample_count': spp  # --> default
+            },
+            'film': {
+                'type': 'hdrfilm',
+                'width': resx, # --> default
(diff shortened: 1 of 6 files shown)


src/libcore/tests/test_xml.py
```diff
@@ -298,3 +298,166 @@ def test20_upgrade_tree(variant_scalar_rgb):
...
+        }
+    writer.configure(scene_dict)
+    s1 = xml.load_dict({
+        'type': 'scene',
+        'sphere':{
+            "type": "sphere",
+            "center" : [0, 0, -10],
+            "radius" : 10.0,
+        }
+        })
```
Speierers, line 322: indentation

src/libcore/tests/test_xml.py
```diff
@@ -298,3 +298,166 @@ def test20_upgrade_tree(variant_scalar_rgb):
...
+    from mitsuba.python.dict_to_xml import WriteXML
+    import os
+    abs_path = os.path.abspath('resources/data/scenes/dict/dict.xml')
+    writer = WriteXML(abs_path)
+    scene_dict = {
+            "type": "sphere",
+            "center" : [0, 0, -10],
+            "radius" : 10.0,
+        }
+    writer.configure(scene_dict)
```
Speierers, line 314: I would rather have a simpler interface, instead of the `WriteXML` class:
```python
abs_path = ...
scene_dict = ...
xml.dict_to_xml(scene_dict, abs_path)
```
What do you think?

src/libcore/tests/test_xml.py
```diff
@@ -298,3 +298,166 @@ def test20_upgrade_tree(variant_scalar_rgb):
...
+        }
+        })
+    s2 = xml.load_file(abs_path)
+    assert str(s1) == str(s2)
+
+@fresolver_append_path
+def test22_xml_missing_type(variant_scalar_rgb):
+    from mitsuba.python.dict_to_xml import WriteXML
+    import os
+    abs_path = os.path.abspath('resources/data/scenes/dict/dict.xml')
```
Speierers, line 330: Could we delete the temporary file once the test finished? Also I wonder if this approach would work when running tests in parallel? We should either have a different file per test, or have the ability to catch the output string (and later use `xml.load_string`).

src/libcore/tests/test_xml.py
Speierers: Also this way the `import` statement would look like:
```python
from mitsuba.python.xml import dict_to_xml
```

src/libcore/tests/test_xml.py
```diff
@@ -298,3 +298,166 @@ def test20_upgrade_tree(variant_scalar_rgb):
...
+    writer.configure(scene_dict)
+    s2 = xml.load_file(abs_path)
+
+    assert str(s1.shapes()[0].bsdf()) == str(s2.shapes()[0].bsdf())
+
+@fresolver_append_path
+def test_24_xml_point(variant_scalar_rgb):
+    from mitsuba.core import xml, Point3f
+    from mitsuba.python.dict_to_xml import WriteXML
+    import os
```
Speierers, line 394: Not sure the `os` is needed in those tests. Maybe you should use Mitsuba's filesystem instead?

src/libcore/tests/test_xml.py
```diff
@@ -298,3 +298,166 @@ def test20_upgrade_tree(variant_scalar_rgb):
                                <float name="intIOR" value="1.33"/>
                            </bsdf>
                        </scene>""")
+
```
Speierers, line 301: We could move those tests in a different test file as this one is pretty large already.

src/libcore/tests/test_xml.py
```diff
@@ -298,3 +298,166 @@ def test20_upgrade_tree(variant_scalar_rgb):
...
+    s1 = xml.load_dict({
+        'type': 'scene',
+        'sphere':{
+            "type": "sphere",
+            "center" : [0, 0, -10],
+            "radius" : 10.0,
+        }
+        })
+    s2 = xml.load_file(abs_path)
+    assert str(s1) == str(s2)
```
Speierers, line 324: It would be great to test the output string directly, the actual XML code produced by the `dict_to_xml` function. Who knows what kind of magic happens in the `xml.load_file`, this extra indirection might hide some implementation bug on the `dict_to_xml` side.

src/python/python/dict_to_xml.py
```diff
@@ -0,0 +1,615 @@
+from collections import OrderedDict
```
Speierers, line 1: Is this necessary?

src/python/python/dict_to_xml.py
```diff
@@ -0,0 +1,615 @@
...
+        Optionally indent the string by a number of tabs
+
+        Returns None
+        '''
+
+        if len(self.files) == 0:
+            scene = object()
+            scene.name = 'untitled'
+            scene.frame_current = 1
+            self.set_filename(scene, 'default')
```
Speierers, line 76: Should this check happen everytime we try to write a line in the files?

src/python/python/dict_to_xml.py
```diff
@@ -0,0 +1,615 @@
...
+            self.set_output_file(file)
+
+        self.wf(self.current_file, '<%s' % name, self.file_tabs[self.current_file])
+
+        for (k, v) in attributes.items():
+            self.wf(self.current_file, ' %s=\"%s\"' % (k, v))
+
+        self.wf(self.current_file, '/>\n')
+
+    def parameter(self, paramType, paramName, attributes={}, file=None):
```
Speierers, line 217: Make sure to use then `underscore_case` naming convention for consistency with the rest of the codebase ;)

src/python/python/dict_to_xml.py
```diff
@@ -0,0 +1,615 @@
...
+
+        params = {
+            'matrix': {
+                'type': 'matrix',
+                'value': value,
+            }
+        }
+        return params
+
+    def decompose_transform(self, transform, export_scale = False):
```
Speierers, line 553: Could we use `enoki` for that?

src/python/python/dict_to_xml.py
```diff
@@ -0,0 +1,615 @@
...
+        }
+        if export_scale:
+            params['scale'] = { #TODO: remove this for cameras
+                'type': 'scale',
+                'value': "%f %f %f" % tuple(scale)
+            }
+
+        return params
+
+def mat_to_euler(mat):
```
Speierers, line 596: This function could probably be useful to other developers. Doesn't it already exist in enoki? Otherwise it should either be added to `enoki` or we could implement this in `C++` and provide bindings for it.

src/libcore/tests/test_xml.py
bathal1: Encapsulating this in a class is quite helpful, as there are numerous operations to do, and buffers to store (for indentation for example). Do you want to change the implementation itself, or simply wrap it in a `dict_to_xml` function that instantiates a `WriteXML` ?

```python
def dict_to_xml(scene_dict, abs_path):
    writer = WriteXML(abs_path)
    writer.configure(scene_dict)
```

src/python/python/dict_to_xml.py
bathal1: This is a relic of the XML writing routine of the old addon that I forgot to remove. I don't think it is needed at all.

src/python/python/dict_to_xml.py
bathal1: oops, good catch

src/libcore/tests/test_xml.py
Speierers: This looks much better!
Also should we rename the method `configure` maybe?

src/libcore/tests/test_xml.py
```diff
@@ -320,31 +320,33 @@ def test21_xml_save_plugin(variant_scalar_rgb):
...
         })
-    s2 = xml.load_file(abs_path)
+    s2 = xml.load_file(filepath)
     assert str(s1) == str(s2)
+    rmtree(os.path.split(filepath)[0])
 
 @fresolver_append_path
 def test22_xml_missing_type(variant_scalar_rgb):
-    from mitsuba.python.dict_to_xml import WriteXML
+    from mitsuba.python.dict_to_xml import dict_to_xml
```
Speierers, line 329: Could we rename the python file?
`dict_to_xml.py -> xml.py`

bathal1: Sure :)

src/libcore/tests/test_xml.py
bathal1: There is now one scene folder per test, removed at the end of the unit test

src/libcore/xml.cpp
```diff
@@ -1174,6 +1174,17 @@ ref<Object> load_string(const std::string &string, const std::string &variant,
...
+    ref<Thread> thread = Thread::thread();
+    ref<FileResolver> fr = thread->file_resolver();
+    ref<FileResolver> fr2 = new FileResolver(*fr);
+    thread->set_file_resolver(fr2);
+
+    // Add the scene file's directory to the search path.
+    fs::path scene_dir = filename_.parent_path();
+    if (!fr2->contains(scene_dir))
+        fr2->append(scene_dir);
+
```
Speierers, line 1187: Could you explain a bit why this is needed?

src/librender/mesh.cpp
```diff
@@ -204,7 +204,11 @@ MTS_VARIANT Mesh<Float, Spectrum>::Mesh(
             // flat shading, use per face normals
             const InputVector3f e1 = face_points[1] - face_points[0];
             const InputVector3f e2 = face_points[2] - face_points[0];
-            normal = normalize(m_to_world.transform_affine(cross(e1, e2)));
+            normal = m_to_world.transform_affine(cross(e1, e2));
+            if(unlikely(all(normal==0.0f)))
```
Speierers, line 208:
```suggestion
            if(unlikely(eq(normal, 0.f)))
```

src/librender/mesh.cpp
```diff
@@ -216,8 +220,11 @@ MTS_VARIANT Mesh<Float, Spectrum>::Mesh(
             Key vert_key;
             if (blender::ME_SMOOTH & face.flag) {
                 // smooth shading, store per vertex normals
-                normal = InputNormal3f(vert.no[0], vert.no[1], vert.no[2]);
-                normal = normalize(m_to_world.transform_affine(normal));
+                normal = m_to_world.transform_affine(InputNormal3f(vert.no[0], vert.no[1], vert.no[2]));
+                if(unlikely(all(normal==0.0f)))
```
Speierers, line 224: same here

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
+import numpy as np
+import os
+from shutil import copy2
+import math
```
Speierers, line 4: let's use `enoki` instead of `math` 😉

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+        'width': 'resx',
+        'height': 'resy'
+    }
+
+    def __init__(self, path, split_files=False):
+        from mitsuba.core import PluginManager
+        self.pmgr = PluginManager.instance()
+        self.split_files = split_files
+        self.scene_data = [{'type': 'scene'}, #MAIN
+                      {}, #MATS
```
Speierers, line 30: indentation

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+import math
+
+class Files:
+    MAIN = 0
+    MATS = 1
+    GEOM = 2
+    EMIT = 3
+    CAMS = 4
+    #TODO: Volumes
+
```
Speierers, line 13: This is a big file, it would be great to add some comments so futur developers can make sense out of it easily.

src/libcore/xml.cpp
bathal1: This is not really related to the `dict_to_xml` implementation, but more of a fix of an issue I encountered when creating unit tests. A given scene XML file may include relative paths to textures, meshes... Therefore we need to add the scene folder to the file resolver, otherwise it won't be able to find these files if we try to load the scene from outside this folder. This was previously done in mitsuba's main function before calling `load_file`, thus ensuring that calling `mitsuba path/to/scene.xml` would work independently of the current working directory. However, this meant that directly calling `load_file` from python would fail, which is why I moved this directly in `load_file`.

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+    '''
+
+    defaults = { #arg name: default name
+        'sample_count': 'spp',
+        'width': 'resx',
+        'height': 'resy'
+    }
+
+    def __init__(self, path, split_files=False):
+        from mitsuba.core import PluginManager
```
Speierers, line 26: Make sure we have set a variant? Without overwriting the current variant (if any) of course

src/python/python/xml.py
Speierers: I am thinking of a general description of what classes and functions in this file can be used for

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+            self.write_header(Files.CAMS, '# Cameras and Render Parameters File')
+
+        self.set_output_file(Files.MAIN)
+
+    def set_output_file(self, file):
+        '''
+        file                int
+
+        Switch next output to the given file index
+
```
Speierers, line 160: Use 
```
Params
-------- 
```
as in the other method descriptions

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+        print('Scene Folder: %s' % self.directory)
+
+        #Set texture directory name
+        self.textures_folder = os.path.join(self.directory, "textures")
+        #create geometry export directory
+        geometry_folder = os.path.join(self.directory, "meshes")
+        if not os.path.isdir(geometry_folder):
+            os.mkdir(geometry_folder)
+
+        #TODO: splitting in different files does not work, fix that
```
Speierers, line 123: Is this fixed now?

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+        else:
+            self.write_header(Files.MAIN)
+
+
+        print('Scene File: %s' % self.file_names[Files.MAIN])
+        print('Scene Folder: %s' % self.directory)
+
+        #Set texture directory name
+        self.textures_folder = os.path.join(self.directory, "textures")
+        #create geometry export directory
```
Speierers, line 118: Minor: comments should start with a space and a capital character 😉

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+                    scene_dict[key] = {'type': 'integer'}
+                elif isinstance(value, float):
+                    scene_dict[key] = {'type': 'float'}
+                elif isinstance(value, str):
+                    scene_dict[key] = {'type': 'string'}
+                elif isinstance(value, bool):
+                    scene_dict[key] = {'type': 'boolean'}
+                else:
+                    raise ValueError("Unsupported default type: %s" % value)
+                #TODO: for now, the only supported defaults are ints, so that works. This may not always be the case though
```
Speierers, line 266: Would be great to support other types as well!

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+                else:
+                    self.data_add(key, value, Files.MAIN)
+            else:
+                raise ValueError("Unsupported item: %s:%s" % (key,value))
+
+
+        #add defaults to MAIN file
+        #self.add_comment("These arguments can be set via command line: -Dargname=value")
+        #for i, default in enumerate(self.defaults.values()):
+        #    self.data_add("__def__%d" % i, {'type': 'default', 'value':default})
```
Speierers, line 321: remove this?

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+                        if args['id'] in self.exported_ids:
+                            raise ValueError("Id: %s is already used!" % args['id'])
+                        self.exported_ids.add(args['id'])
+                    if len(value) > 0: #open a tag if there is still stuff to write
+                        self.open_element(tag, args)
+                        self.write_dict(value)
+                        self.close_element()
+                    else:
+                        self.element(tag, args) #write dict in one line (e.g. integrator)
+                #elif value_type in {'scale', 'rotate', 'translate', 'matrix', 'ref', 'default', 'include', 'integer', 'string', 'boolean', 'float'}:
```
Speierers, line 462: .

src/python/python/xml.py
bathal1: Well the question is more: which defaults do we automatically add? For now, the defaults added are all ints

src/python/python/xml.py
bathal1: oops, yes

src/libcore/xml.cpp
Speierers: I think previously those lines where written in the python script explicitly. See the [doc](https://mitsuba2.readthedocs.io/en/latest/src/python_interface/parsing_xml.html#loading-a-scene):
```
# Absolute or relative path to the XML file
filename = 'path/to/my/scene.xml'

# Add the scene directory to the FileResolver's search path
Thread.thread().file_resolver().append(os.path.dirname(filename))

# Load the scene for an XML file
scene = load_file(filename)
```

src/libcore/tests/test_write_xml.py
```diff
@@ -0,0 +1,213 @@
+import enoki as ek
+import pytest
+import mitsuba
+
+from mitsuba.python.test.util import fresolver_append_path
+
+@fresolver_append_path
+def test21_xml_save_plugin(variant_scalar_rgb):
```
Speierers, line 8: Tests should start at `01`

src/libcore/tests/test_write_xml.py
```diff
@@ -0,0 +1,213 @@
...
+    from mitsuba.core import xml
+    from mitsuba.python.xml import dict_to_xml
+    import os
+    from shutil import rmtree
```
Speierers, line 12: import those two once at the top

src/python/python/xml.py
```diff
@@ -0,0 +1,617 @@
...
+            }
+        }
+        return params
+
+    def decompose_transform(self, transform, export_scale = False):
+        '''
+        Export a transform as a combination of rotation, scale and translation.
+        This helps manually modifying the transform after export (for cameras for instance)
+        '''
+        transform = transform.matrix.numpy()
```
Speierers, line 556: Using enoki `transform_decompose` for this?

src/libcore/xml.cpp
bathal1: Why not do it directly in `load_file` then?

src/libcore/tests/test_write_xml.py
```diff
@@ -0,0 +1,301 @@
...
+
+from mitsuba.python.test.util import fresolver_append_path
+
+@fresolver_append_path
+def test01_xml_save_plugin(variant_scalar_rgb):
+    from mitsuba.core import xml
+    from mitsuba.core import Thread
+    from mitsuba.python.xml import dict_to_xml
+    fr = Thread.thread().file_resolver()
+    filepath = os.path.join(str(fr[len(fr)-1]), 'resources/data/scenes/dict01/dict.xml')
```
Speierers, line 15: Maybe add a comment here to explain what this is doing ^^

src/librender/mesh.cpp
```diff
@@ -181,8 +185,11 @@ MTS_VARIANT Mesh<Float, Spectrum>::Mesh(
             Key vert_key;
             if (blender::ME_SMOOTH & face.flag) {
                 // smooth shading, store per vertex normals
-                normal = InputNormal3f(vert.no[0], vert.no[1], vert.no[2]);
-                normal = normalize(m_to_world.transform_affine(normal));
+                normal = m_to_world.transform_affine(InputNormal3f(vert.no[0], vert.no[1], vert.no[2]));
+                if(unlikely(all(eq(normal, 0.f))))
+                    fail("Mesh has invalid normals!");
+                else
+                    normal = normalize(normal);
```
Speierers, line 192:
```suggestion
                if(unlikely(all(eq(normal, 0.f))))
                    Throw("Mesh has invalid normals!");
                
                normal = normalize(normal);
```

PR #134, opened 2020-05-12
Minor fixes to radiancemeter plugin

leroyvn:
This small PR fixes a bug in the `radiancemeter` plugin where an omitted ray update would lead to invalid intersections (it is the same as in #121, I forgot to update `sample_ray_differential` in that PR).

In addition, I set the `m_needs_sample_2` and `m_needs_sample_3` members to `false`.

Edit: I also fixed wrong `version` tags in the `irradiancemeter` plugin tests.


Speierers:
Thanks for this @leroyvn, I will merge this PR.

PR #135, opened 2020-05-12
Implicit shapes support with Optix

Speierers:
This PR adds support for custom shapes (e.g. `Sphere`, `Disk`, ...) to the OptiX backend.

### Codebase structure changes

- `/mitsuba/include/render/optix/` contains optix/cuda header files used by the OptiX backend

- `/mitubsa/src/librender/optix/optix_rt.cu` implemented the OptiX programs and includes custom shape CUDA files. PTX code can be compiled using the `/mitsuba/resources/ptx/MakeFile`

- `/mitsuba/src/shapes/optix/` contains the custom shape optix implementation

- `/mitsuba/include/render/optix/shapes.h` needs to be modified when adding a new custom shape to OptiX backend. 

### Add a custom shape

Shape plugin need to provide a CUDA header file with the implementation of two OptiX programs:

- `__intersection__myplugin()`: computes the ray intersection
- `__closesthit__myplugin()`: computes and fills the surface interaction output data

as well as a struct defining the shape data needed in the programs above. (e.g. `OptixMyShapeData`).

Then, the C++ plugin needs to overwrite the `optix_prepare_geometry()` method to be supported by the optix backend. This method allocate a GPU buffer and uploads the data needed in the `__intersection__` and `__closesthit__` programs. For instance, the `Disk` program only needs to upload it's `to_world` and `to_object` transforms:
```c++
#if defined(MTS_ENABLE_OPTIX)
    using Base::m_optix_data;

    void optix_prepare_geometry() override {
        if constexpr (is_cuda_array_v<Float>) {
            Base::optix_prepare_geometry();
            // Only allocates the first time
            if (m_optix_data.empty()) {
                m_optix_data = empty<DynamicBuffer<Int8>>(sizeof(OptixDiskData));
                m_optix_data.managed();
            }

            store(m_optix_data.data() + offsetof(OptixDiskData, to_world),  m_to_world);
            store(m_optix_data.data() + offsetof(OptixDiskData, to_object), m_to_object);
        }
    }
#endif
```

Note that the plugin can access the `OptixDistData` struct by including the corresponding CUDA header files, for instance for `Disk`:
```c++
    #include "optix/disk.cuh"
```


src/shapes/optix/mesh.cuh
```diff
@@ -0,0 +1,94 @@
...
+struct OptixMeshData {
+    const optix::Vector3u *faces;
+    const optix::Vector3f *vertex_positions;
+    const optix::Vector3f *vertex_normals;
+    const optix::Vector2f *vertex_texcoords;
+};
+
+#ifdef __CUDACC__
+
+/// Read in an array of Array<T,3> with a true stride of 3 (and not 4, see
```
wjakob, line 14: Is this still relevant?

include/mitsuba/render/optix/shapes.h
```diff
@@ -0,0 +1,35 @@
...
+#include "cylinder.cuh"
+#include "disk.cuh"
+#include "mesh.cuh"
+#include "rectangle.cuh"
+#include "sphere.cuh"
+#else
+NAMESPACE_BEGIN(mitsuba)
+/// List of custom OptiX shape desciptors
+/// (ClassName, ClosesthitFuncName, IntersectionFuncName)
+static std::tuple<const char*, const char*, const char*> custom_optix_shapes[] = {
```
wjakob, line 15: Taking another look at this, it seems like a very "hardcoded" way to do string concatenation :-). I would propose that we allow the shape names to be upper case in those method names, and then we can just use the normal "+" operator of `std::string` to add the prefixes.

src/shapes/optix/mesh.cuh
Speierers: Yes as the mesh data buffers (e.g. `vertex_positions`) are stored without alignment (3 floats per point and not 4 floats).

include/mitsuba/render/optix/shapes.h
Speierers: Actually I originally tried to concatenate those strings as you are suggesting. Somehow the optix routine that is using them (`optixProgramGroupCreate`) kept segfaulting when forging one of those string manually. I assumed it requires those `const char *` to be in the data segment somehow. (really not sure about this)

PR #138, opened 2020-05-14
Use a smaller render block sizes when appropriate

tizian:
Automatically decrease the render block size if it produces fewer blocks
than available render threads.


wjakob:
Thanks!

Issue #127, opened 2020-05-06
Flexible spectral sampling for sensors


leroyvn:
This is a discussion about how to make Mitsuba's spectral sampling more flexible (follow-up to another discussion with @wjakob). I already got some guidance on how to proceed and I'm sharing here a few additional thoughts.

# Targeted features

1. Defining at runtime the spectral range on which the computation will be performed. This is useful for many applications not directly related to rendering.
2. Sampling wavelengths in exotic subsets of the covered spectral range (_e.g._ union of disjoint intervals, set of discrete values). This feature would not be of much use without proper post-processing machinery but it's off the scope of this specific discussion.
3. Varying wavelength weighting to allow for the simulation of the sensor's spectral response (please correct me if I'm wrong).

# Proposed implementation

All of this seems achievable by leveraging the current spectral sampling facilities with minor modifications. The biggest change would be that sensors supporting this flexible sampling system would not use `sample_wavelength` directly anymore but would instead rely on a spectrum plugin, very much like what the `point` plugin's `sample_ray` does.

We'd then have a series of spectrum plugins to handle that. In order to not mess with scene definition (I assume most users actually don't care about that and would like not to hear about it), a default plugin implementing the current spectral sampling method would be provided.

If this seems like a good idea, I'll prepare a PR when I'll be done with all that.

wjakob:
Hi Vincent,

just some quick feedback: I would welcome features to provide a custom spectral range and wavelength weighting at runtime. Before you implement it, we should discuss a bit what could be a good mechanism to specify such information (maybe starting with XML fragments, then moving to the API level).

Best,
Wenzel

leroyvn:
Thanks a lot for answering so quickly! Here are a couple of examples illustrating what I have in mind.

Let's assume that `perspective` has been modified to support flexible spectrum sampling. The  sensor's spectral response function (SRF, there is probably better wording for a graphics-related context) is specified using a new parameter `srf`. In the following example, the spectrum is uniformly sampled between 400 and 700 nm:
```xml
<sensor version="2.0.0" type="perspective">
  <spectrum name="srf" type="srf_uniform">
    <float name="lambda_min" value="400"/>
    <float name="lambda_max" value="700"/>
  </spectrum>
</sensor>
```

In the following, wavelengths are sampled from a set of discrete values:
```xml
<sensor version="2.0.0" type="perspective">
  <spectrum name="srf" type="srf_discrete">
[...]

PR #142, opened 2020-05-19
Flexible spectral sampling

leroyvn:
This PR adds flexible spectral sampling for sensors as discussed in #127. Changes are as follows. I only added it to the ` perspective` sensor plugin so far, I'll add this to the other sensor plugins depending on feedback.

## Perspective camera plugin

I added a new member `m_srf` which, if set, defines a spectrum from which sensor wavelengths will be sampled. As requested by @wjakob, I tried to minimize the performance overhead in default cases. This means that:

- if unspecified, `m_srf` will be set to `nullptr`;
- if `m_srf` is set when using a non-spectral variant, it will be ignored, set to `nullptr` and a warning will be issued;
- if `m_srf` is set, it will be used to sample wavelengths when calling `sample_ray` or `sample_ray_differential`;
- otherwise, the default RGB sampling method will be used without a virtual function call.

Tests have been updated accordingly.

## Uniform spectrum plugin

I added the possiblity to specify a spectral interval for which the uniform value is applied; otherwise, the plugin returns 0. This led to adding two plugin parameters `lambda_min` and `lambda_max`, as well as the corresponding data members. All parameters are optional, including `value`, which now defaults to 1 (not sure if it's desirable behaviour, I just thought it would be convenient sometimes). I'm however unsure about the performance cost brought by the multiple parameter checks I introduced: is it okay?

## Example renders

Default Cornell box:

cbox_00

Constant SSR (value=1) in [400, 500 nm] using modified `uniform` spectrum plugin:

cbox_01

Sentinel-2A/MSI Band 03 SSR using a loaded `irregular` plugin:

cbox

## Todo

- Should I add render tests?
- [ ] Add support for all sensor plugins
- [ ] Update spectrum documentation
- [ ] Update sensor documentation


src/spectra/tests/test_uniform.py
```diff
@@ -0,0 +1,73 @@
+import pytest
+
+import enoki as ek
+import mitsuba
+
+
+def make_spectrum(value=None, lambda_min=None, lambda_max=None):
+    from mitsuba.core.xml import load_string
```
Speierers, line 8: Would be great to use `load_dict` instead of `load_string` in the unit tests. This enables better readibility.

You can have a look at the doc [here](https://mitsuba2.readthedocs.io/en/latest/src/python_interface/parsing_xml.html#creating-objects-using-python-dictionaries)

The plan is to slowly rewrite all the tests so that they use `load_dict`.

leroyvn:
Hi @Speierers, I pushed an update to the `uniform` plugin tests. I tried refactoring those for `perspective`, but I get (for some tests, not all of them), errors like:
```
RuntimeError: ​[xml_v.cpp:259] Unkown value type: <class 'mitsuba.core.Transform4f'>
```

This is the refactored sensor creation function:
```python
def create_camera(o, d, fov=34, fov_axis="x", s_open=1.5, s_close=5, dict_srf=None):
    from mitsuba.core.xml import load_dict
    from mitsuba.core import Transform4f, Vector3f

    dict_camera = {
        "type": "perspective",
        "near_clip": 1.0,
        "far_clip": 35.0,
        "focus_distance": 15.0,
        "to_world": Transform4f.look_at(
            origin=Vector3f(o),
            target=Vector3f(o) + Vector3f(d),
            up=[0, 1, 0]
        ),
        "fov": fov,
        "fov_axis": fov_axis,
        "shutter_open": s_open,
        "shutter_close": s_close,
        "film": {
            "type": "hdrfilm",
            "width": 512,
            "height": 256,
        }
    }

    if dict_srf:
        dict_camera["srf"] = dict_srf

    return load_dict(dict_camera)
```

Speierers:
> Any idea of what could be wrong?

Maybe you should try using `ScalarTransform4f`?

leroyvn:
Still no luck using `ScalarTransform4f`. I however noticed that this error occurs with several variants, I'll create an issue for that.

leroyvn:
Hi @Speierers, I'm reviving this PR a little: would you mind telling me if this implementation suits you? I could then proceed with propagating this to the other sensor plugins.

I was actually wondering if it wouldn't be better (for maintenance) to have this implemented as part of `Sensor`, but given what @wjakob said about minimising virtual function calls, I'm afraid this would impact performance. Are my doubts justified?

leroyvn:
Hi @wjakob, I just finished propagating the sensor response function implementation to all the currently available sensor plugins (including relevant docs updates). The currently open questions are:
- is this code suitable for merge?
- render tests: should I add a couple of render tests based on the examples I showed in the PR text?
- extended spectral range tutorial: should I add a docs page (I guess in the "Advanced topics" section) explaining how to modify the spectral range?

Speierers:
Hi @leroyvn and @wjakob ,

Should we revive this PR and continue working on this in prospect to merging this into `next`?

leroyvn:
Hi @Speierers, I think this is still relevant to atmospheric radiative transfer use cases (although we're still not using it). I however wonder if changes could not be made more concise with some `Sensor` (or maybe `Endpoint`?) interface updates.

PR #144, opened 2020-05-20
BlenderMesh Plugin

bathal1:
As requested by @Speierers , Blender meshes are now loaded using a separate plugin, `BlenderMesh`


src/librender/mesh.cpp
```diff
@@ -228,13 +228,13 @@ MTS_VARIANT Mesh<Float, Spectrum>::Mesh(
                 map_entry->value   = vert_id;
                 map_entry->is_init = true;
                 // add stuff to the temporary buffers
-                tmp_vertices.push_back(
-                    m_to_world.transform_affine(face_points[i]));
-                tmp_normals.push_back(normal);
+                InputPoint3f pt = m_to_world.transform_affine(face_points[i]);
+                tmp_vertices.push_back({pt.x(), pt.y(), pt.z()});
```
Speierers, line 232: This should be a bit more efficient:
```suggestion
                tmp_vertices.emplace_back(pt.x(), pt.y(), pt.z()};
```

wjakob: If your data structure is POD (plain old data) and doesn't need fancy bits like a move constructor, the compiler will optimize both of these variants to the same assembly (though I also prefer the `emplace_back` for readability).

What is super-important though, and I didn't see that here, is to call ``std::vector::reserve`` up-front to reserve the right amount of memory for the number of vertices that you plan to add. It can be a conservative estimate as well in case you don't now 100%. If you start with a 0-sized vector, you'll have to pay with a whole bunch of unnecessary copies as the implementation is progressively resizing to a larger and larger buffer.

bathal1: Thank you @wjakob for your feedback. Depending on the mesh's connectivity and shading options, the number of stored unique vertices/normals/UVs can be severely reduced, so it may be difficult to have good initial guess, though we could reserve `vertex_count` vertices, since it can't be more reduced than that.

src/shapes/blender.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "blender_types.h"
+
+NAMESPACE_BEGIN(mitsuba)
+
+/**
+
+Blender mesh loader
+----------------------------------------------------------
+
+This plugins converts a Blender Mesh to mitsuba's mesh layout
```
Speierers, line 12: Could be more descriptive on how to use this in Python? E.g. could provide a simple code snippet that would be great!

src/shapes/blender.cpp
```diff
@@ -0,0 +1,263 @@
+#include <mitsuba/render/mesh.h>
+#include <enoki/color.h>
+#include "blender_types.h"
```
Speierers, line 3: If this isn't used anywhere else, maybe could move the code of this header in this file?

src/shapes/blender.cpp
```diff
@@ -0,0 +1,263 @@
+#include <mitsuba/render/mesh.h>
+#include <enoki/color.h>
```
Speierers, line 2: Do we need this?

bathal1: yes, for `srgb_to_linear`

Speierers:
Thanks 🚀

PR #154, opened 2020-05-26
Fix initialization and shutdown of optix to be done globally

dvicini:
Fixes the problem that optix was initialized and shutdown on a per scene basis


Speierers:
Awesome, looks good to me, go ahead an merge it.

PR #156, opened 2020-05-28
Update rendering_scene.rst

matthewpurri:
The Bitmap module wasn't imported so converting the film to jpeg failed.


Speierers:
Thanks!

PR #165, opened 2020-06-04
XML resource path

Speierers:
This PR adds a new XML tag `resource_path` to add path to the file resolver's list of search paths directly from an XML scene file.

Using this new feature, the `resource/data` repo became a lot cleaner, with the introduction of a `common` folder, containing meshes and textures used for testing and documentation images generation, ...

### DIFF
--- a/docs/src/getting_started/file_format.rst
+++ b/docs/src/getting_started/file_format.rst
@@ -407,7 +407,6 @@ precedence when no command line arguments are given. The syntax for this is:
 
 and must precede the occurrences of the parameter in the XML file.
 
-
 Including external files
 ------------------------
 
@@ -444,3 +443,14 @@ can be accomplished using the ``alias as=".."`` tag:
 After this statement, the diffuse scattering model will be bound to *both*
 identifiers ``my_material_1`` and ``my_material_2``.
 
+External resource folders
+-------------------------
+
+Using the ``path`` tag, it is possible to add a path to the list of search paths. This can
+be useful for instance when some meshes and textures are stored in a different directory, (e.g. when
+shared with other scenes). The path can be absolute, relative to the directory containing the XML
+scene file, or relative to any path already existing in the list of search paths.
+
+.. code-block:: xml
+
+    <path value="../../my_resources"/>
--- a/docs/src/getting_started/file_format.rst
+++ b/docs/src/getting_started/file_format.rst
@@ -448,8 +448,9 @@ External resource folders
 
 Using the ``path`` tag, it is possible to add a path to the list of search paths. This can
 be useful for instance when some meshes and textures are stored in a different directory, (e.g. when
-shared with other scenes). The path can be absolute, relative to the directory containing the XML
-scene file, or relative to any path already existing in the list of search paths.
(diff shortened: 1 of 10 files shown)


wjakob:
The implementation added some code to the recursive tree traversal where we enter a node, but there is no cleanup. This means that scene loading will actually modify the `FileResolver` of the caller, potentially accumulating large numbers of paths over time. You will need to remove the added path at the end, perhaps when we are done parsing the root element. Would be good to have a test case for this too, which loads a scene and checks that the FileResolver is preserved. Alternatively you could clone the FileResolver before starting to load the scene and set back the original one later on. Caution: scene loading can fail, so it will need to be guarded by some kind of `try {} catch (…) { throw; }` contraption.

I don’t think calling `fs->resolve` on the supplied path is the best design — this function will recursively try to find the path in every location that’s already pushed onto the `FileResolver`, which is going to be hard to understand if there are a few paths in there already. I propose a simpler interpretation: absolute paths are pushed directly, and that relative paths are expressed relative to the directory containing the XML file currently being parsed. This should be realizable in a cross-platform manner using the already existing `fs::*` api.

Finally, bike shedding — I can’t warm up to the “resourcepath” tag name which IMO doesn’t look as nice as all of the other short 1-word tag names in the scene description language. Resource also sounds to me like textures or shapes, but it could in fact be a full scene.

How about simply:

```
<path value=“”/>
```
(reversing my prior comment to not take up that name)

Finally… append or prepend, that is the question. I think makes more sense to prepend. Why? Imagine loading an XML file that includes another XML file. We want to have the added paths of the recursively included file to take precedence over those of the parent XML file.

Right now, paths can be added at any point in the file (even e.g. as the attribute of a bitmap class :-)). That could be quite confusing. How about checking that they are children of the root element and erroring out otherwise?

Speierers:
@wjakob, another tag name could be `folder`:
```
<folder value="" />
```

merlinND:
It could also be `<include dir="">` (which doesn't clash with `<include filename="">`), or `<dir value="">`.

I agree that pre-pending makes a lot of sense. We could also have an attribute `prepend` which defaults to true, but can be overriden.

Speierers:
Personally, I would prefer having a seperate tag (e.g. `path`, `dir`, ...) rather than using the `include` tag.

Speierers:
Another thing, when using the `load_string` function, there is not actual XML file. Therefore, I would suggest that for relative path, we first try to "resolve" it relatively to the XML file (which will always fail for `load_string`) and then try to resolve it relatively to the other path present in the FileResolver.

Or we can just accept the fact that for `load_string`, one need to give absolute path only.

Speierers:
@wjakob thanks for your feedbacks. I implemented those in the last commits.

I will squash those once I get the green light for merging ;)

wjakob:
This now looks good to me, thank you @Speierers. Just one minor suggestion: in the documentation, can you clarify how the path will be resolved? E.g.: if it is a relative path, Mitsuba 2 will first try to interpret it relative to the scene directory, then other paths that are already on the search path (you can also mention the command line argument).

By the way: Mitsuba 1 IIRC had a command line argument for the `mitsuba` binary to add search paths explicitly. Would be nice to have that as well and then mention it in the documentation as well.

PR #170, opened 2020-06-08
Instancing support (`shapegroup` and `instance` plugins)

Speierers:
This is PR ports the `shapegroup` and `instance` plugins to Mitsuba 2.

Thanks @nathan96g for his great work!

## TODOs

- [x] Minor fixes in current implementation
- [x] Check Embree support
- [x] OptiX support
- [x] Documentation
- [ ] Add more tests


include/mitsuba/render/optix/ray.cuh
```diff
@@ -0,0 +1,41 @@
...
+
+namespace optix {
+
+struct Ray3f {
+    Vector3f o;       ///< Ray origin
+    Vector3f d;       ///< Ray direction
+    Vector3f d_rcp;   ///< Componentwise reciprocals of the ray direction
+    float mint;     ///< Minimum position on the ray segment
+    float maxt;     ///< Maximum position on the ray segment
+    float time;     ///< Time value associated with this ray
```
Speierers, line 15: align those comments maybe?

include/mitsuba/render/optix/ray.cuh
```diff
@@ -0,0 +1,41 @@
...
+        : o(o), d(d), d_rcp(frcp(d)), mint(mint), maxt(maxt), time(time) { }
+
+    /// Return the position of a point along the ray
+    DEVICE Vector3f operator() (float t) const { return fmaf(t, d, o); }
+
+    /// Update the reciprocal ray directions after changing 'd'
+    DEVICE void update() { d_rcp = frcp(d); }
+};
+
+DEVICE Ray3f get_ray_instance_space() {
```
Speierers, line 30: Worth adding a comment here. Also could we rename this function? `get_object_ray()` maybe?

include/mitsuba/render/optix/shapes.h
```diff
@@ -27,5 +32,213 @@ size_t get_shape_descr_idx(Shape *shape) {
...
+};
+
+template <typename Shape>
+void fill_hitgroup_records(std::vector<HitGroupSbtRecord> &hitgroup_records,
+                           std::vector<ref<Shape>> &shapes,
+                           OptixProgramGroup *program_groups) {
+    for (size_t i = 0; i < 2; i++)
+        for (Shape* shape: shapes)
+            if (i == !shape->is_mesh())
+                shape->optix_fill_hitgroup_records(hitgroup_records, program_groups);
```
Speierers, line 58: Let's add the braces expect for the `if` statement here. Also maybe add a comment that meshes a processed first.

include/mitsuba/render/optix/shapes.h
```diff
@@ -27,5 +32,213 @@ size_t get_shape_descr_idx(Shape *shape) {
...
+        build_input.instanceArray.instances = (CUdeviceptr) d_instances;
+        build_input.instanceArray.numInstances = instances.size();
+        build_input.instanceArray.aabbs = 0;
+        build_input.instanceArray.numAabbs = 0;
+
+        OptixAccelBufferSizes buffer_sizes;
+        rt_check(optixAccelComputeMemoryUsage(context, &accel_options, &build_input, 1, &buffer_sizes));
+
+        void* d_temp_buffer = cuda_malloc(buffer_sizes.tempSizeInBytes);
+        accel.buffer_ias  = cuda_malloc(buffer_sizes.outputSizeInBytes);
```
Speierers, line 221: extra space here?

src/shapes/instance.cpp
```diff
@@ -0,0 +1,201 @@
...
+            return instance;
+        } else {
+            Throw("embree_geometry() should only be called in CPU mode.");
+        }
+    }
+#endif
+
+#if defined(MTS_ENABLE_OPTIX)
+    virtual void optix_prepare_instance(const OptixDeviceContext& context, OptixInstance& instance, uint32_t instance_id) override {
+        // TODO: check coeffs
```
Speierers, line 176: Could remove this todo?

src/shapes/instance.cpp
```diff
@@ -0,0 +1,201 @@
...
+        std::memcpy(instance.transform, transform, sizeof(float) * 12);
+        instance.instanceId = instance_id;
+        instance.visibilityMask = 255;
+        instance.flags = OPTIX_INSTANCE_FLAG_NONE;
+        m_shapegroup->optix_accel_handle(context, instance.traversableHandle, instance.sbtOffset);
+    }
+
+    virtual void optix_fill_hitgroup_records(std::vector<HitGroupSbtRecord>&, OptixProgramGroup*) override {
+        /* no op */
+    }
```
Speierers, line 191: Do we need really need to override this method?

4str0m: Yes, because the base implementation appends a hitgroup, but an instance doesn't need any:
```
MTS_VARIANT void Shape<Float, Spectrum>::optix_fill_hitgroup_records(std::vector<HitGroupSbtRecord> &hitgroup_records, OptixProgramGroup *program_groups) {
    optix_prepare_geometry();
    // Set hitgroup record data
    hitgroup_records.push_back(HitGroupSbtRecord());
    hitgroup_records.back().data = { (uintptr_t) this, m_optix_data_ptr };

    size_t program_group_idx = (is_mesh() ? 2 : 3 + get_shape_descr_idx(this));
    // Setup the hitgroup record and copy it to the hitgroup records array
    rt_check(optixSbtRecordPackHeader(program_groups[program_group_idx], &hitgroup_records.back()));
}
```

include/mitsuba/render/optix/matrix.cuh
```diff
@@ -124,6 +125,21 @@ struct Transform4f {
 #else
     DEVICE Transform4f() { }
 
+    DEVICE Transform4f(float m[12], float inv[12])
+    {
```
wjakob, line 129: indentation

include/mitsuba/render/optix/shapes.h
```diff
@@ -27,5 +32,217 @@ size_t get_shape_descr_idx(Shape *shape) {
...
+            buffer_sizes.tempSizeInBytes,
+            (CUdeviceptr)accel.buffer_ias,
+            buffer_sizes.outputSizeInBytes,
+            &accel.handle,
+            0,  // emitted property list
+            0   // num emitted properties
+        ));
+
+        cuda_free((void*)d_temp_buffer);
+        cuda_free((void*)d_instances); // TODO: check if we can free this now...
```
wjakob, line 243: Comment still warranted?

include/mitsuba/render/optix/vector.cuh
```diff
@@ -231,6 +231,14 @@ DEVICE Array<Value, Size> fmaf(const Value &a, const Array<Value, Size> &b, cons
         result.v[i] = ::fmaf(a, b[i], c[i]);
     return result;
 }
+
+template <typename Value, size_t Size>
+DEVICE Array<Value, Size> frcp(const Array<Value, Size> &a) {
+    Array<Value, Size> result;
+    for (size_t i = 0; i < Size; ++i)
+        result.v[i] = 1.f / a[i];
```
wjakob, line 239: `__frcp_rn`?

include/mitsuba/render/optix/vector.cuh
```diff
@@ -231,6 +231,14 @@ DEVICE Array<Value, Size> fmaf(const Value &a, const Array<Value, Size> &b, cons
         result.v[i] = ::fmaf(a, b[i], c[i]);
```
wjakob, line 231: __fmaf_rn

src/librender/scene_embree.inl
```diff
@@ -28,14 +28,22 @@ MTS_VARIANT void Scene<Float, Spectrum>::accel_init_cpu(const Properties &/*prop
 
     Timer timer;
     RTCScene embree_scene = rtcNewScene(__embree_device);
+    rtcSetSceneFlags(embree_scene,RTC_SCENE_FLAG_DYNAMIC);
```
wjakob, line 31: space after comma

src/librender/scene_embree.inl
```diff
@@ -65,20 +73,30 @@ Scene<Float, Spectrum>::ray_intersect_cpu(const Ray3f &ray, Mask active) const {
...
-                si.shape = m_shapes[shape_index];
                 si.prim_index = prim_index;
 
-                // Create the cache for the Mesh shape
-                Float cache[2] = { rh.hit.u, rh.hit.v };
-
-                // Ask shape to fill in the rest
-                si.shape->fill_surface_interaction(ray, cache, si);
+                // If the hit is not on an instance
+                if( inst_index == RTC_INVALID_GEOMETRY_ID ) {
```
wjakob, line 87: spacing: `if<space>(<no space>expr<no space>)`

src/librender/scene_embree.inl
```diff
@@ -65,20 +73,30 @@ Scene<Float, Spectrum>::ray_intersect_cpu(const Ray3f &ray, Mask active) const {
...
+                    // If the hit is not on an instance
+                    si.shape = m_shapes[shape_index];
+                    // Create the cache for the Mesh shape
+                    Float cache[2] = { rh.hit.u, rh.hit.v };
+                    // Ask shape to fill in the rest
+                    si.shape->fill_surface_interaction(ray, cache, si);
+                } else {
+                    // If the hit is on an instance
+                    si.instance = m_shapes[inst_index];
```
wjakob, line 96: `si.shape` becomes undefined/`NULL` on this branch?

IIRC, In Mitsuba 1: `shape` is *always* the intersected shape, and `instance` can contain a pointer to the `ShapeGroup` if applicable.

src/shapes/shapegroup.cpp
```diff
@@ -0,0 +1,251 @@
...
+#endif
+
+        m_shapegroup = true;
+    }
+
+#if defined(MTS_ENABLE_EMBREE)
+    void init_embree_scene(RTCDevice device) override {
+        if constexpr (!is_cuda_array_v<Float>) {
+            // Construct the BVH only once
+            if(m_scene == nullptr) {
```
wjakob, line 125: spacing.

src/shapes/shapegroup.cpp
```diff
@@ -0,0 +1,251 @@
...
+    void release_embree_scene() override {
+        if constexpr (!is_cuda_array_v<Float>) {
+            rtcReleaseScene(m_scene);
+        }
+    }
+
+    RTCGeometry embree_geometry(RTCDevice device) const override {
+        if constexpr (!is_cuda_array_v<Float>) {
+            RTCGeometry instance = rtcNewGeometry(device, RTC_GEOMETRY_TYPE_INSTANCE);
+            if(m_scene == nullptr)
```
wjakob, line 143: spacing

src/shapes/shapegroup.cpp
```diff
@@ -0,0 +1,251 @@
...
+        std::ostringstream oss;
+            oss << "ShapeGroup[" << std::endl
+                << "  name = \"" << m_id << "\"," << std::endl
+                << "  prim_count = " << primitive_count() << std::endl
+                << "]";
+        return oss.str();
+    }
+
+#if defined(MTS_ENABLE_OPTIX)
+    virtual void optix_accel_handle(const OptixDeviceContext& context, OptixTraversableHandle &out_handle, uint32_t &out_sbt_offset) override {
```
wjakob, line 218: wrap long line?

src/shapes/shapegroup.cpp
```diff
@@ -0,0 +1,251 @@
...
+        if (m_accel.handle == 0ull)
+            build_gas(m_shapes, context, m_sbt_offset, m_accel);
+        out_handle = m_accel.handle;
+        out_sbt_offset = m_sbt_offset;
+    }
+
+    virtual void optix_fill_hitgroup_records(std::vector<HitGroupSbtRecord> &hitgroup_records, OptixProgramGroup *program_groups) override {
```
wjakob, line 225: ditto

include/mitsuba/render/shape.h
```diff
@@ -327,6 +329,12 @@ class MTS_EXPORT_RENDER Shape : public Object {
     /// Is this shape a triangle mesh?
     bool is_mesh() const { return m_mesh; }
```
wjakob, line 330: The `Shape` class has racked up quite a few of these fields that add  storage to every Mitsuba shape. We already have this information via the `m_class` member. I would be in favor of removing all of those fields and just writing `m_class->derives_from(MTS_CLASS(Mesh))` and `m_class->name() == "Instance"` and `m_class->name() == "ShapeGroup"` in the other two case. AFAIK none of these functions are actually called during rendering (just scene initialization), so this should be fine. (but I could be wrong)

4str0m:
> Another thing to consider: sometimes such methods are extremely specific to a single plugin like “instance” or “shapegroup” — check out Mitsuba 1. There, the instance actually had its own header file (int the “shape” directory) to add a few function declarations that were not in `shape.h`. I propose that we follow a similar approach.

This method was usefull for cases like the `optix_accel_handle` method, which was only defined inside the `ShapeGroup` plugin, and used in the `Instance` plugin. Unfortunately, even though it could be done for the `optix_prepare_instance` method (that is currently only defined inside the `Instance` plugin), this method is used in `librender` (more specifically in the `include/mitsuba/render/optix/shapes.h` file). This prevents it to be defined in an `instance.h` header, as the actual implementation would have to be visible to the linker when compiling `librender.so`.

@wjakob : do you have an idea on how we could keep the current architecture (`Instance` and `Shapegroup` classes as separate plugins) while still allowing  the `optix_prepare_instance` method to be specific to the instance's plugin only.

wjakob:
@4strom: thinking about this a bit more, I think it would only be possible if you declare these as inline method in a hypothetical `instance.h`/`shapegroup.h` that is then included called from `librender`. That is the only way to avoid having to link `librender` to the plugin. Effectively it would mean that `librender` knows about the instance layout of these two classes and then performs any method calls directly there. What do you think?

wjakob:
(This, by the way, is IIRC also how it was handled in Mitsuba 1.)

PR #173, opened 2020-06-09
Fix envmap sampling and wrapping

Speierers:
This patch fixes the `envmap` evalutation/interpolation routine so that it matches the one of the underlying `Distribution2D` used for sampling.

`Distribution2D` expects a 2D weight field in which the values of the first row/column match the values of the last row/column. As this is not the case in regular HDRi files, the constructor of envmap adds one row and column to the image data in order to satisfy this.

### DIFF
--- a/src/emitters/envmap.cpp
+++ b/src/emitters/envmap.cpp
@@ -79,34 +79,24 @@ public:
         bitmap = bitmap->convert(Bitmap::PixelFormat::RGBA, struct_type_v<ScalarFloat>, false);
         m_filename = file_path.filename().string();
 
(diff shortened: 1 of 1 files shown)


baranusta:
This code seems to be introducing another bug.

The bug can be reproduced at a scene with a mirror sphere. An environment map whose pixels are black produces a red spot on the sphere in the top direction of where the last(or first) column of env map is. 

image

Speierers:
Hi @baranusta,

Thanks for reporting this issue. It is fixed in the upcoming new codebase, hence let me close this PR.

PR #184, opened 2020-06-19
Allocate one extra triangle in Mesh buffers

Speierers:
This patch makes sure that the `Mesh` buffers allocate for one extra triangle.

This is necessary as `enoki::Array<float, 3>` is actually represented as a `enoki::Array<float, 4>` (e.g. better alignment) under the hood. This causes out of memory accessing when reading the very last triangle in the buffers (read one extra float).

Unfortunately, this could be confusing when looking at the `slices(buffer)` as it won't match `3*vertex_count` anymore. @wjakob is there another way to allocate an extra bit of memory while keeping the slice the same?


wjakob:
Let's keep this PR in limbo for now, it might be a moot point with the new JIT modes.

PR #187, opened 2020-06-22
Sampler plugins

Speierers:
## Implementation of various `sampler` plugins

This PR provides implementation for various sampler plugins:

- `stratified`
- `multijitter` (from [Correlated Multi-jitter sampling](https://graphics.pixar.com/library/MultiJitteredSampling/))
- `ldsampler` (ported from Mitsuba 1)
- `orthogonal` (from [Orthogonal Array Sampling for Monte Carlo Rendering](https://cs.dartmouth.edu/~wjarosz/publications/jarosz19orthogonal.html))

This PR also update the doc API.

### DIFF
--- a/include/mitsuba/core/qmc.h
+++ b/include/mitsuba/core/qmc.h
@@ -183,4 +183,47 @@ private:
     int m_scramble;
 };
 
+
+/// Van der Corput radical inverse in base 2
+template <typename UInt, typename Float = float_array_t<UInt>>
+Float radical_inverse_2(UInt index, UInt scramble = 0) {
+    if constexpr (is_double_v<Float>) {
+        index = (index << 32) | (index >> 32);
+        index = ((index & 0x0000ffff0000ffffULL) << 16) | ((index & 0xffff0000ffff0000ULL) >> 16);
+        index = ((index & 0x00ff00ff00ff00ffULL) << 8)  | ((index & 0xff00ff00ff00ff00ULL) >> 8);
+        index = ((index & 0x0f0f0f0f0f0f0f0fULL) << 4)  | ((index & 0xf0f0f0f0f0f0f0f0ULL) >> 4);
+        index = ((index & 0x3333333333333333ULL) << 2)  | ((index & 0xccccccccccccccccULL) >> 2);
+        index = ((index & 0x5555555555555555ULL) << 1)  | ((index & 0xaaaaaaaaaaaaaaaaULL) >> 1);
+
+        /* Generate an uniformly distributed double precision number in [1,2)
+         * from the scrambled index and subtract 1. */
+        return reinterpret_array<Float>(sr<12>(index ^ scramble) | 0x3ff0000000000000ull) - 1.0;
+    } else {
+        index = (index << 16) | (index >> 16);
+        index = ((index & 0x00ff00ff) << 8) | ((index & 0xff00ff00) >> 8);
+        index = ((index & 0x0f0f0f0f) << 4) | ((index & 0xf0f0f0f0) >> 4);
+        index = ((index & 0x33333333) << 2) | ((index & 0xcccccccc) >> 2);
+        index = ((index & 0x55555555) << 1) | ((index & 0xaaaaaaaa) >> 1);
+
+        /* Generate an uniformly distributed single precision number in [1,2)
+         * from the scrambled index and subtract 1. */
+        return reinterpret_array<Float>(sr<9>(index ^ scramble) | 0x3f800000u) - 1.f;
+    }
+}
+
+
+/// Sobol' radical inverse in base 2
+template <typename UInt, typename Float = float_array_t<UInt>>
+Float sobol_2(UInt index, UInt scramble = 0) {
+    if constexpr (is_double_v<Float>) {
+        for (UInt v = 1ULL << 52; index != 0; index >>= 1, v ^= v >> 1)
+            masked(scramble, eq(index & 1U, 1U)) ^= v;
+        return reinterpret_array<Float>(sr<12>(scramble) | 0x3ff0000000000000ull) - 1.0;
+    } else {
+        for (UInt v = 1U << 31; index != 0; index >>= 1, v ^= v >> 1)
+            masked(scramble, eq(index & 1U, 1U)) ^= v;
+        return Float(scramble) / Float(1ULL << 32);
+    }
+}
+
 NAMESPACE_END(mitsuba)
--- a/include/mitsuba/core/random.h
+++ b/include/mitsuba/core/random.h
@@ -28,9 +28,11 @@
 
 #pragma once
 
+#include <mitsuba/core/math.h>
 #include <mitsuba/core/simd.h>
 #include <mitsuba/core/logger.h>
 #include <mitsuba/core/traits.h>
+#include <mitsuba/core/fwd.h>
 #include <enoki/random.h>
 
 NAMESPACE_BEGIN(enoki)
@@ -169,6 +171,106 @@ auto sample_tea_float(UInt v0, UInt v1, int rounds = 4) {
         return sample_tea_float32(v0, v1, rounds);
     else
         return sample_tea_float64(v0, v1, rounds);
+}
+
+/**
+ * \brief Generate pseudorandom permutation vector using a shuffling network and the
+ * \ref sample_tea function. This algorithm has a O(log2(sample_count)) complexity but
+ * only supports permutation vectors whose lengths are a power of 2.
+ *
+ * \param index
+ *     Input index to be mapped
+ * \param sample_count
+ *     Length of the permutation vector
+ * \param seed
+ *     Seed value used as second input to the Tiny Encryption Algorithm. Can be used to
+ *     generate different permutation vectors.
+ * \param rounds
+ *     How many rounds should be executed by the Tiny Encryption Algorithm? The default is 2.
+ * \return
+ *     The index corresponding to the input index in the pseudorandom permutation vector.
+ */
+
+template <typename UInt32>
+UInt32 permute(UInt32 index, uint32_t sample_count, UInt32 seed, int rounds = 2) {
+    uint32_t  n = log2i(sample_count);
+    Assert((1 << n) == sample_count, "sample_count should be a power of 2");
+
+    for (uint32_t  level = 0; level < n; ++level) {
+        UInt32 bit = UInt32(1 << level);
+        // Take a random integer indentical for indices that might be swapped at this level
+        UInt32 rand = sample_tea_32(index | bit, seed, rounds);
+        masked(index, eq(rand & bit, bit)) = index ^ bit;
+    }
+
+    return index;
+}
+
+/**
+ * \brief Generate pseudorandom permutation vector using the algorithm described in Pixar's
+ * technical memo "Correlated Multi-Jittered Sampling":
+ *
+ *     https://graphics.pixar.com/library/MultiJitteredSampling/
+ *
+ *  Unlike \ref permute, this function supports permutation vectors of any length.
+ *
+ * \param index
+ *     Input index to be mapped
+ * \param sample_count
+ *     Length of the permutation vector
+ * \param seed
+ *     Seed value used as second input to the Tiny Encryption Algorithm. Can be used to
+ *     generate different permutation vectors.
+ * \return
+ *     The index corresponding to the input index in the pseudorandom permutation vector.
+ */
+
+template <typename UInt32>
+UInt32 permute_kensler(UInt32 index, uint32_t sample_count, UInt32 seed, mask_t<UInt32> active = true) {
+    if (sample_count == 1)
+        return zero<UInt32>(slices(index));
+
+    UInt32 w = sample_count - 1;
+    w |= w >> 1;
+    w |= w >> 2;
+    w |= w >> 4;
+    w |= w >> 8;
+    w |= w >> 16;
+
+    // Worst case is when the index is sequentially mapped to every invalid numbers (out
+    // of range) before being mapped into the correct range. E.g. decreasing sequence
+    uint32_t max_iter = 0;
+    if constexpr (is_cuda_array_v<UInt32>)
+        max_iter = math::round_to_power_of_two(sample_count) - sample_count + 1;
+
+    uint32_t iter = 0;
+    mask_t<UInt32> invalid = true;
+    UInt32 tmp;
+    do {
+        tmp = index;
+        tmp ^= seed;
+        tmp *= 0xe170893d;
+        tmp ^= seed >> 16;
+        tmp ^= (tmp & w) >> 4;
+        tmp ^= seed >> 8;
+        tmp *= 0x0929eb3f;
+        tmp ^= seed >> 23;
+        tmp ^= (tmp & w) >> 1;
+        tmp *= 1 | seed >> 27;
+        tmp *= 0x6935fa69;
+        tmp ^= (tmp & w) >> 11;
+        tmp *= 0x74dcb303;
+        tmp ^= (tmp & w) >> 2;
+        tmp *= 0x9e501cc3;
+        tmp ^= (tmp & w) >> 2;
+        tmp *= 0xc860a3df;
+        tmp &= w;
+        tmp ^= tmp >> 5;
+        masked(index, invalid) = tmp;
+        invalid = (index >= sample_count);
+    } while (any_or<false>(active && invalid) || (max_iter > ++iter));
+
+    return (index + seed) % sample_count;
 }
 
 NAMESPACE_END(mitsuba)
--- a/include/mitsuba/render/sampler.h
+++ b/include/mitsuba/render/sampler.h
@@ -8,13 +8,14 @@
 #include <mitsuba/core/random.h>
 
 NAMESPACE_BEGIN(mitsuba)
+
 /**
  * \brief Base class of all sample generators.
  *
  * For each sample in a pixel, a sample generator produces a (hypothetical)
  * point in the infinite dimensional random number cube. A rendering
  * algorithm can then request subsequent 1D or 2D components of this point
- * using the \ref next_1d() and \ref next_2d() functions.
+ * using the \c next_1d and \c next_2d functions.
  *
  * Scalar and wavefront rendering algorithms will need interact with the sampler
  * interface in a slightly different way:
@@ -73,7 +74,7 @@ public:
     /**
      * \brief Advance to the next sample.
      *
-     * A subsequent call to \ref next_1d() or \ref next_2d() will access the first
+     * A subsequent call to \c next_1d or \c next_2d will access the first
      * 1D or 2D components of this sample.
      */
     virtual void advance();
--- a/include/mitsuba/render/sampler.h
+++ b/include/mitsuba/render/sampler.h
@@ -5,9 +5,45 @@
 #include <mitsuba/core/logger.h>
 #include <mitsuba/core/object.h>
 #include <mitsuba/core/vector.h>
+#include <mitsuba/core/random.h>
 
 NAMESPACE_BEGIN(mitsuba)
-
+/**
+ * \brief Base class of all sample generators.
+ *
+ * For each sample in a pixel, a sample generator produces a (hypothetical)
+ * point in the infinite dimensional random number cube. A rendering
+ * algorithm can then request subsequent 1D or 2D components of this point
+ * using the \ref next_1d() and \ref next_2d() functions.
+ *
+ * Scalar and wavefront rendering algorithms will need interact with the sampler
+ * interface in a slightly different way:
+ *
+ * Scalar rendering algorithm:
+ *
+ *   1. Before beginning to render a pixel block, the rendering algorithm calls
+ *      \c seed to initialize a new sequence with the specific seed offset.
+ *   2. The first pixel sample can now be computed, after which \c advance needs to be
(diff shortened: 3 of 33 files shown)


include/mitsuba/render/sampler.h
```diff
@@ -41,20 +88,54 @@ class MTS_EXPORT_RENDER Sampler : public Object {
...
+    /// Set the number of samples per pass in wavefront modes (default is 1)
+    void set_samples_per_wavefront(uint32_t samples_per_wavefront);
 
     MTS_DECLARE_CLASS()
 protected:
     Sampler(const Properties &props);
     virtual ~Sampler();
 
+    /// Generates a array of seeds where the seed values are unique per sequence
+    UInt32 compute_per_sequence_seed(uint64_t seed_offset) const;
```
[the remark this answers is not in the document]
wjakob: (Or maybe the argument should be an uint32_t? In any case, that difference seems strange.)

include/mitsuba/render/sampler.h
```diff
@@ -41,20 +88,54 @@ class MTS_EXPORT_RENDER Sampler : public Object {
...
+    /// Return the index of the current sample
+    UInt32 current_sample_index() const;
+
+protected:
+    uint64_t m_base_seed;
+    uint32_t m_sample_count;
+    uint32_t m_samples_per_wavefront;
+    /// Array of offsets to compute current sample index in wavefront modes
+    UInt32 m_wavefront_sample_offsets;
```
wjakob, line 114: Is there any fundamental reason that this needs to be stored as a variable? In practice, this computation should be merged into a kernel and then recomputed at every wavefront. Paying the cost of reading a variable just to avoid a division or modulo is not a good trade-off for the GPU.

src/samplers/independent.cpp
```diff
@@ -28,63 +28,43 @@ by Melissa O’Neill.
 
 This is the most basic sample generator; because no precautions are taken to avoid
 sample clumping, images produced using this plugin will usually take longer to converge.
-In theory, this sampler is initialized using a deterministic procedure, which means
-that subsequent runs of Mitsuba should create the same image. In practice, when
-rendering with multiple threads and/or machines, this is not true anymore, since the
-ordering of samples is influenced by the operating system scheduler.
+this sampler is initialized using a deterministic procedure, which means that subsequent runs
+of Mitsuba should create the same image.
```
wjakob, line 32: Not sure that the removal of the previous text is 100% correct. When rendering things in parallel on the CPU, the order in which blocks are merged into the film *will* cause very small differences. You could add a note that these should be absolutely negligible, with relative errors on the order of the machine epsilon (:math:`6\cdot 10^{-8}`) in single precision.

src/samplers/independent.cpp
```diff
@@ -28,63 +28,43 @@ by Melissa O’Neill.
...
+
+.. subfigstart::
+.. subfigure:: ../../resources/data/docs/images/sampler/independent_1024_samples.svg
+   :caption: 1024 samples projected onto the first two dimensions.
+.. subfigure:: ../../resources/data/docs/images/sampler/independent_64_samples_and_proj.svg
+   :caption: 64 samples projected onto the first two dimensions and their
+             projection on both 1D axis (top and right plot). Both the 2D and 1D projections
+             exhibit sample clumping which will result in higher variance in the rendered image.
```
wjakob, line 40: The casual reader might not know what "sample clumping" really implies. I would move that text from the figure caption to the main documentation block above and explain that there are both regions that don't receive many samples (i.e. we don't know much about the behavior of the function there), and regions where many samples are very close together (which likely have very similar values).

src/samplers/independent.cpp
```diff
@@ -28,63 +28,43 @@ by Melissa O’Neill.
...
-            m_rng->seed(sample_tea_64(seed_value, idx),
-                        sample_tea_64(idx, seed_value));
-        } else {
-            m_rng->seed(seed_value, PCG32_DEFAULT_STREAM + arange<UInt64>());
-        }
-    }
-
     Float next_1d(Mask active = true) override {
-        if constexpr (is_dynamic_array_v<Float>) {
-            if (m_rng == nullptr)
```
wjakob, line 78: Are these kinds of error checks still somewhere in the code? They were quite helpful to debug sampler-related bugs.

src/samplers/ldsampler.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+.. _sampler-ldsampler:
+
+Low discrepancy sampler (:monosp:`ldsampler`)
+---------------------------------------------
+
+This plugin implements a simple hybrid sampler that combines aspects of a Quasi-Monte Carlo sequence
+with a pseudorandom number generator based on a technique proposed by Kollig and Keller
+:cite:`Kollig2002Efficient`. It is a good and fast general-purpose sample generator. There exists
+other QMC samplers that can generate even better distributed samples, but this comes at a higher
```
wjakob, line 19: There exists other -> Other QMC samplers exist that [..]

src/samplers/ldsampler.cpp
```diff
@@ -0,0 +1,135 @@
...
+cost in terms of performance. This plugin is based on Mitsuba 1's default sampler (also called
+:monosp:`ldsampler`).
+
+Roughly, the idea of this sampler is that all of the individual 2D sample dimensions are first
+filled using the same (0, 2)-sequence, which is then randomly scrambled and permuted using a
+shuffle network. The name of this plugin stems from the fact that (0, 2) sequences minimize the
+so-called star disrepancy, which is a quality criterion on their spatial distribution.
```
wjakob, line 26: Instead of "so-called", better to link to some page with more details (e.g. https://en.wikipedia.org/wiki/Low-discrepancy_sequence)

wjakob: While the (0,2) sequence clearly has a very low star discrepancy, I'm not sure if "minimizes" is the correct way of introducing this QMC sequence. (Probably only asymptotically? It's been a long time since I worked on these, and I don't remember 100%..). Anyways, would be good to have a mathematically sound description here, I'd look at how it is introduced in the relevant literature.

src/samplers/ldsampler.cpp
```diff
@@ -0,0 +1,135 @@
...
+.. subfigend::
+   :label: fig-ldsampler-renders
+
+.. subfigstart::
+.. subfigure:: ../../resources/data/docs/images/sampler/ldsampler_1024_samples.svg
+   :caption: 1024 samples projected onto the first two dimensions.
+.. subfigure:: ../../resources/data/docs/images/sampler/ldsampler_64_samples_and_proj.svg
+   :caption: A projection of the first 64 samples onto the first two dimensions and their
+             projection on both 1D axis (top and right plot). The 1D stratification is perfect as
+             this sampler doesn't add additional jittering.
```
wjakob, line 42: Wouldn't use "jittering" in text, it is not a clearly defined term.

src/samplers/ldsampler.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    void seed(uint64_t seed_offset, size_t wavefront_size) override {
+        Base::seed(seed_offset, wavefront_size);
+        m_scramble_seed = compute_per_sequence_seed(seed_offset);
+    }
+
+    Float next_1d(Mask active = true) override {
+        ENOKI_MARK_USED(active);
+
+        UInt32 sample_indices = current_sample_index();
```
wjakob, line 94: The lines of this function could use a little more documentation.

src/samplers/ldsampler.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    Point2f next_2d(Mask active = true) override {
+        ENOKI_MARK_USED(active);
+
+        UInt32 sample_indices = current_sample_index();
+        UInt32 perm_seed = m_scramble_seed + m_dimension_index++;
+
+        UInt32 i = permute(sample_indices, m_sample_count, perm_seed);
+        UInt32 scramble_x = sample_tea_32(m_scramble_seed, UInt32(0x18bc21eb));
+        UInt32 scramble_y = sample_tea_32(m_scramble_seed, UInt32(0x58633e2d));
```
wjakob, line 111: Ditto, also where are these constants coming from?

src/samplers/multijitter.cpp
```diff
@@ -0,0 +1,159 @@
...
+.. _sampler-multijitter:
+
+Correlated Multi-Jittered sampler (:monosp:`multijitter`)
+---------------------------------------------------------
+
+.. pluginparameters::
+
+ * - sample_count
+   - |int|
+   - Number of samples per pixel. This value might be slightly increased by the plugin to produce a
```
wjakob, line 19: `This value might be slightly ` -> `The sampler may internally choose to slightly increase [..] to create a subdivision into strata that has an aspect ratio close to one`

src/samplers/multijitter.cpp
```diff
@@ -0,0 +1,159 @@
...
+     strata grid with an aspect ratio close to one. (Default: 4)
+ * - seed
+   - |int|
+   - Seed offset (Default: 0)
+ * - jitter
+   - |bool|
+   - Adds additional random jitter withing the substratum (Default: True)
+
+This plugin implements the methods introduced in Pixar's tech memo :cite:`kensler1967correlated`.
```
wjakob, line 28: IIRC this was a paper in graphics gems, would be nice to cite the peer-reviewed version. If the tech memo adds further detail, then you can add a second citation.

src/samplers/multijitter.cpp
```diff
@@ -0,0 +1,159 @@
...
+    void seed(uint64_t seed_offset, size_t wavefront_size) override {
+        Base::seed(seed_offset, wavefront_size);
+        m_permutation_seed = compute_per_sequence_seed(seed_offset);
+    }
+
+    Float next_1d(Mask active = true) override {
+        UInt32 sample_indices = current_sample_index();
+        UInt32 perm_seed = m_permutation_seed + m_dimension_index++;
+
+        Float p = kensler_permute(sample_indices, m_sample_count, perm_seed * 0x45fbe943, active);
```
wjakob, line 106: Needs documentation for each line of the crucial bits.

src/samplers/orthogonal.cpp
```diff
@@ -0,0 +1,225 @@
...
+   - Orthogonal array's strength (Default: 2)
+ * - seed
+   - |int|
+   - Seed offset (Default: 0)
+ * - jitter
+   - |bool|
+   - Adds additional random jitter withing the substratum (Default: True)
+
+This plugin implements the Orthogonal Array sampler generator introduced by
+:cite:`jarosz19orthogonal`. It generalizes correlated multi-jittered sampling to higher dimensions
```
wjakob, line 31: Technical writing tip: never use a citation as a word in a sentence. It is metadata that is not pronounced when reading the text. Better: `Jaroz et al. [citation link]`

src/samplers/orthogonal.cpp
```diff
@@ -0,0 +1,225 @@
...
+by using *orthogonal arrays (OAs)* of strength :math:`s`. The strength property of OAs tells that
```
wjakob, line 32: An OA of strength `s` has the property that ..

src/samplers/orthogonal.cpp
```diff
@@ -0,0 +1,225 @@
...
+projecting the generated samples to any combination of :math:`s` dimensions will always result in
+a well stratified pattern. In other words, when :math:`s=2` (default value), the high-dimentional
+samples are simultaneously stratified in all 2D projections as if they had been produced by
+correlated multi-jittered sampling. By construction, samples produced by this generator are also
+well stratified when projected on both 1D axis.
```
wjakob, line 37: Why wouldn't one choose `s=1000` here to achieve the best stratification over many dimensions? Needs a discussion of the pros/cons.

src/samplers/orthogonal.cpp
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /// Bush construction technique for orthogonal arrays
+    Float bush(UInt32 i,       // sample index
+               ScalarUInt32 j, // dimension
+               ScalarUInt32 s, // number of levels/stratas (has to be a prime number)
+               ScalarUInt32 t, // strength of OA
+               UInt32 p,       // pseudo-random permutation seed
+               Mask active = true) {
+        ScalarUInt32 N = enoki::pow(s, t);
```
wjakob, line 163: `ScalarUInt32` -> `uint32_t` everywhere.

Normally `pow(a, b) = exp(log(a) * b)`, which would be a very bad idea. There is a special case somewhere in Enoki that specializes to integer exponent. It would be good to double-check that that is executed here..

src/samplers/orthogonal.cpp
```diff
@@ -0,0 +1,225 @@
...
+        i = kensler_permute(i, N, p, active);
+        auto i_digits = to_base_s(i, s, t);
+        ScalarUInt32 stm = N / s;
+        UInt32 phi = eval_poly(i_digits, j);
```
wjakob, line 167: Line explaining what each step does.

src/samplers/orthogonal.cpp
```diff
@@ -0,0 +1,225 @@
...
+    Float bose(UInt32 i,       // sample index
+               ScalarUInt32 j, // dimension
+               ScalarUInt32 s, // number of levels/stratas (has to be a prime number)
+               UInt32 p,       // pseudo-random permutation seed
+               Mask active = true) {
+
+        // Permutes the sample index so that samples are obtained in random order
+        i = kensler_permute(i % (s * s), s * s, p, active);
+
+        // TODO could use enoki::divisor here
```
wjakob, line 187: Please do :).

src/samplers/orthogonal.cpp
```diff
@@ -0,0 +1,225 @@
...
+        UInt32 a_i0 = i / s;
+        UInt32 a_i1 = i % s;
+
+        UInt32 a_ij, a_ik;
```
wjakob, line 191: Comments briefly explaining the different steps here.

include/mitsuba/render/sampler.h
```diff
@@ -5,9 +5,48 @@
...
+ * point in the infinite dimensional random number cube. A rendering
+ * algorithm can then request subsequent 1D or 2D components of this point
+ * using the \c next_1d and \c next_2d functions.
+ *
+ * Scalar and wavefront rendering algorithms will need interact with the sampler
+ * interface in a slightly different way:
+ *
+ * Scalar rendering algorithm:
+ *
+ *   1. Before beginning to render a pixel, the rendering algorithm calls
```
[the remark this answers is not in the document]
Speierers: It is true in the current implementation. E.g. every pixel renders with a different seed offset. This is for scalar modes only thought.

src/librender/sampler.cpp
```diff
@@ -1,23 +1,108 @@
...
+
+    m_dimension_index = 0u;
+    m_sample_index = 0;
+}
 
-MTS_VARIANT Float Sampler<Float, Spectrum>::next_1d(Mask) { NotImplementedError("next_1d"); }
+MTS_VARIANT void Sampler<Float, Spectrum>::advance() {
+    m_dimension_index = 0u;
+    m_sample_index++;
+    Assert(m_sample_index < (m_sample_count / m_samples_per_wavefront));
```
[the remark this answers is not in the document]
Speierers: All of those variables are scalar values so it should be fine.

include/mitsuba/render/sampler.h
Speierers: I will replace it with `uint32_t m_wavefront_size` then.

src/samplers/independent.cpp
Speierers: I see, we are talking about numerical precision uncertainties in this case right?

src/samplers/independent.cpp
Speierers: `m_rng` isn't a pointer anymore so I removed that check. If you would like to add a assertion whether the sampler was seeded, we can check that `m_wavefront_size != 0`.

include/mitsuba/render/sampler.h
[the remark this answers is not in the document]
wjakob: That seems excessive, how about seeding once per block? (At least for scalar mode)

src/samplers/independent.cpp
wjakob: I wouldn't call it uncertainty, just minor variation due to OS scheduling in the presence of arithmetic that is non-commutative due to builtin rounding.

src/samplers/independent.cpp
wjakob: Yeah, that sounds good.

src/samplers/multijitter.cpp
wjakob: Ah got it, I was just confused.

include/mitsuba/core/qmc.h
```diff
@@ -183,4 +183,50 @@ class MTS_EXPORT_CORE RadicalInverse : public Object {
...
+        index = ((index & 0x0000ffff0000ffffULL) << 16) | ((index & 0xffff0000ffff0000ULL) >> 16);
+        index = ((index & 0x00ff00ff00ff00ffULL) << 8)  | ((index & 0xff00ff00ff00ff00ULL) >> 8);
+        index = ((index & 0x0f0f0f0f0f0f0f0fULL) << 4)  | ((index & 0xf0f0f0f0f0f0f0f0ULL) >> 4);
+        index = ((index & 0x3333333333333333ULL) << 2)  | ((index & 0xccccccccccccccccULL) >> 2);
+        index = ((index & 0x5555555555555555ULL) << 1)  | ((index & 0xaaaaaaaaaaaaaaaaULL) >> 1);
+
+        // Account for the available precision and scramble
+        index = (index >> (64 - 53)) ^ (scramble & ~-(1LL << 53));
+
+        return Float(index) / Float(1ULL << 53);
```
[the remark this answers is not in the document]
Speierers: Something like this I suppose?
```c++
        index = (index << 32) | (index >> 32);
        index = ((index & 0x0000ffff0000ffffULL) << 16) | ((index & 0xffff0000ffff0000ULL) >> 16);
        index = ((index & 0x00ff00ff00ff00ffULL) << 8)  | ((index & 0xff00ff00ff00ff00ULL) >> 8);
        index = ((index & 0x0f0f0f0f0f0f0f0fULL) << 4)  | ((index & 0xf0f0f0f0f0f0f0f0ULL) >> 4);
        index = ((index & 0x3333333333333333ULL) << 2)  | ((index & 0xccccccccccccccccULL) >> 2);
        index = ((index & 0x5555555555555555ULL) << 1)  | ((index & 0xaaaaaaaaaaaaaaaaULL) >> 1);

        /* Generate an uniformly distributed double precision number in [1,2)
         * from the scrambled index and subtract 1. */
        return reinterpret_array<Float64>(sl<20>(index ^ scramble) |
                                          0x3ff0000000000000ull) - 1.0;
```

include/mitsuba/core/random.h
```diff
@@ -171,4 +172,97 @@ auto sample_tea_float(UInt v0, UInt v1, int rounds = 4) {
...
+        tmp *= 0x74dcb303;
+        tmp ^= (tmp & w) >> 2;
+        tmp *= 0x9e501cc3;
+        tmp ^= (tmp & w) >> 2;
+        tmp *= 0xc860a3df;
+        tmp &= w;
+        tmp ^= tmp >> 5;
+        masked(index, invalid) = tmp;
+        invalid = (index >= sample_count);
+    } while (any(active && invalid));
```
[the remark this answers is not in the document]
Speierers: In the worst case this will run `A-B` iterations, where `B` is the permutation vector size (here `sample_count`) and `A` is the next power of two after `B`. So if `sample_count = 100` this should run at most 28 times.

include/mitsuba/core/random.h
```diff
@@ -171,4 +172,97 @@ auto sample_tea_float(UInt v0, UInt v1, int rounds = 4) {
...
+        tmp = index;
+        tmp ^= seed;
+        tmp *= 0xe170893d;
+        tmp ^= seed >> 16;
+        tmp ^= (tmp & w) >> 4;
+        tmp ^= seed >> 8;
+        tmp *= 0x0929eb3f;
+        tmp ^= seed >> 23;
+        tmp ^= (tmp & w) >> 1;
+        tmp *= 1 | seed >> 27;
```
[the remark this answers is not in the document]
Speierers: Not enterly sure as this code was generated by a "genetic algorithm". This is how it is described in both Pixar's and Jarosv's papers tho.

include/mitsuba/render/sampler.h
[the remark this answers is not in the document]
Speierers: You are right, this is what's happening, it is called once per block. I will update the doc.

src/samplers/ldsampler.cpp
Speierers: Those are arbitrary constants. We use them here to get different scramble values along the x and y axis from the same scramble seed.

src/samplers/orthogonal.cpp
Speierers: This isn't how it works. That is the difference between Orthogonal Arrays and Strong Orthogonal Arrays. For OA, the stratification will be good only for the s-tuple of dimensions and for the 1D projections. E.g. when `s=3`, there is no garantee that stratification hold in 2D.

Maybe this wasn't clear in the comment, I will try to add a few words on this.

PR #200, opened 2020-07-06
Add defaults to all occurences of a parameter.

bathal1:
This PR fixes [this](https://github.com/mitsuba-renderer/mitsuba2-blender/issues/8) bug in the blender addon.


Speierers:
Thanks for this @bathal1 .

Could you please squash those commits into a single commit?

Also would it be possible to add a unit test for this?

Speierers:
Thanks!

PR #201, opened 2020-07-06
Bumpmap bsdf adapter plugin basic implementation

ommaury:
Here's a basic implementation for a bump map bsdf adapter plugin, I've updated the code to align its style on the recently committed normal map plugin.

**Known limitations**

1. No filtering: neither through basic texture filtering (not supported in mts2 yet) nor through more sophisticated NDF-based techniques.

2. No validity checks performed on resulting wo directions, which leads to the well-known black regions if the normal perturbation is too strong. If you're interested, we could talk about implementing something like:
    - what iray or cycles do, i.e. on-the-fly normal corrections to make sure the reflection vectors are always above the surface. In my experience, these solutions are quite effective, but they suffer as you know from being (even more) non-symmetrical.
    - microfacet-based normal mapping to keep the solution symmetrical and energy conserving: https://cg.ivd.kit.edu/publications/2017/normalmaps/normalmap.pdf.

3. Support for uv-based gradient evaluation only (could be extended to support position-based gradient).

**Supporting scene files**

Here are the 2 xml files for the docs:

bsdf_bumpmap_without.xml:
```
<scene version="2.1.0">
    <path value="../../common"/>

    <bsdf type="roughplastic" id="object_bsdf">
        <texture name="diffuse_reflectance" type="bitmap">
            <string name="filename" value="textures/floor_tiles.jpg"/>
        </texture>
    </bsdf>

    <include filename="include/matpreview.xml"/>
</scene>
```

bsdf_bumpmap_with.xml:
```
<scene version="2.1.0">
    <path value="../../common"/>

    <bsdf type="bumpmap" id="object_bsdf">
        <float name="strength" value="0.02"/>
        <texture name="bumpmap" type="bitmap">
            <boolean name="raw" value="true"/>
            <string name="filename" value="textures/floor_tiles_bumpmap.png"/>
        </texture>
        <bsdf type="roughplastic">
            <texture name="diffuse_reflectance" type="bitmap">
                <string name="filename" value="textures/floor_tiles.jpg"/>
            </texture>
        </bsdf>
	</bsdf>

    <include filename="include/matpreview.xml"/>
</scene>
```
Here's the `floor_tiles_bumpmap.png` texture (I used the recently checked-in floor tiles normal map to compute it):

floor_tiles_bumpmap

And this is what the render should look like:

bsdf_bumpmap_with


wjakob:
Hi @ommaury,

thank you for your PR. I think that this would be a great feature to have in Mitsuba, but there are some architectural aspects that I would personally like to see solved in a different way.

1. The decision to depend on differentials strikes me as unusual -- I think that it would be preferable to have a purely local definition of the model that only relies on perfect analytic derivatives, and which behaves in an identical way at all bounces.

2. Instead of finite differences, I would propose to extend the Texture interface with differential versions of the various evaluation routines. Something like `eval_3_grad()`, which returns both the function value and an UV gradient.

3. To perform bump mapping, all that is needed is a Frame3f, and that is cheaper to construct than a complete SurfaceInteraction3f data structure (which is unnecessarily large). Check out the tentative normal map plugin that was committed today (src/bsdf/normalmap.cpp).

Thanks,
Wenzel

ommaury:
Hi Wenzel,

I have just committed another pass taking into account your comments. The `Texture` interface extension is currently limited to `eval_1_grad()` and `eval_3_grad()` and `Bitmap` only implements `eval_1_grad()` since that seemed like the only case used in practice: I'm not sure how exhaustive you want the specializations to be.

For the docs, the xml above will generate the same image :)

Thanks,
Olivier

wjakob:
Thank you, that looks like a great improvement. I left some low level comments regarding masking in the code. I also have three more high-level requests, and then it’s good to go:


1. while the SurfaceInteraction3f data structure provides world-space shading frame (“sh_frame”) and surface normal (“n”), shading is purely implemented in local coordinates. Sh_frame is used to convert to this space and is not accessed in BSDF::eval() and friends. So various world-to-local and back conversions are not needed, and it’s preferable that BSDF code is purely expressed in a frame where the surface normal is (0, 0, 1). You can see these kinds of simplifications here: https://github.com/mitsuba-renderer/mitsuba2/commit/e4df3ff28b65e133660718ede3ba32f40dd58ec6. Basically one can just apply the small rotation that perturbs the frame, instead of involving world space at all.

2. The conventions of this plugin are a little different from the bump map plugin in Mitsuba 1, which called “strength” “scale”, and which accepted any nested Texture no matter its name. Unless there are strong reasons for changing this, it would be nice to retain compatibility.

3. It would be nice to have at least some kind of basic testcase to ensure the continued functionality of this plugin as we go through refactoring etc. in the future. This could involve querying instantiating a perturbed diffuse BRDF from Python and querying the value for some angle. Or you could send us your scene (is the texture image ok for us to use license-wise) and we’ll include it in the test suite. Or both.

Thanks!
Wenzel

ommaury:
Hi Wenzel,

1. Great insight, thanks :) Having removed the world transforms in the rest of the code, I'm currently converting the perturbed normal computed in world space back to local before returning, it could arguably be done directly in local space, projecting `dpdu` and `dpdv` onto that shading frame, up to you.

2. I've renamed `strength` to `scale`, and made the texture used for bump a nested `Texture` plugin, which can have now an arbitrary name.

3.  You can use the scene above for the doc and test, it is basically the same as the one you have created for testing the normalmap plugin. The 2k bump texture `floor_tiles_bumpmap.png` (above) is your normal map texture converted to a height field, feel free to use it.

```
<scene version="2.1.0">
    <path value="../../common"/>

    <bsdf type="bumpmap" id="object_bsdf">
        <float name="strength" value="0.02"/>
        <texture name="floor_tiles_bumpmap" type="bitmap">
            <boolean name="raw" value="true"/>
            <string name="filename" value="textures/floor_tiles_bumpmap.png"/>
        </texture>
        <bsdf type="roughplastic">
            <texture name="diffuse_reflectance" type="bitmap">
                <string name="filename" value="textures/floor_tiles.jpg"/>
            </texture>
        </bsdf>
	</bsdf>

    <include filename="include/matpreview.xml"/>
</scene>
```

4. I've realized the gradient implementation is currently incomplete so I added a TODO comment: if the underlying uv layout transform involves a rotation for example, the bitmap (x,y) and (u,v) layout are not aligned and the code doesn't take that into account. I can do another pass later if you want, unless you already have an idea of how you want to handle that.

5. Concerning masking, I'm a bit confused:
  - when is it a good idea to take it into account?
  - when is it required to take it into account?

My current understanding is that, thanks to the `ENOKI_CALL_SUPPORT_METHOD` wrapping mechanism, the `BSDF` methods such as `sample()`, `eval()` and `pdf()` will gather the appropriate portions of the result arrays for each unique bsdf pointer and re-assemble full arrays, overwriting in the process potential unwanted temporary results from unmasked calculations: masking inside these calls does not seem strictly required then, and rather an optimization question. Is that correct?

Thanks,
Olivier

ommaury:
Hi Wenzel,

I went ahead and:
- refactored and extended the `Texture::eval_1_grad()` to support arbitrary uv transforms. I added its python binding and a unit test with supporting file `noise_8x8.png` (https://github.com/mitsuba-renderer/mitsuba-data/pull/7) to test the analytic gradient vs. finite difference on random uv rotations. I added it `src/textures/tests/test_bitmap.py`.
- removed `Texture::eval_3_grad()`, would suggest adding back if/when you actually need it.
- added all the bumpmap supporting doc/test scene files and corresponding images:
  https://github.com/mitsuba-renderer/mitsuba-data/pull/7

Hope that helps,
Olivier

src/bsdfs/bumpmap.cpp
```diff
@@ -0,0 +1,210 @@
...
+                                             Mask active) const override {
+        MTS_MASKED_FUNCTION(ProfilerPhase::BSDFSample, active);
+
+        // Sample nested BSDF with perturbed shading frame
+        SurfaceInteraction3f perturbed_si(si);
+        perturbed_si.sh_frame = frame(si, active);
+        perturbed_si.wi = perturbed_si.to_local(si.wi);
+        auto [bs, weight] = m_nested_bsdf->sample(ctx, perturbed_si,
+                                                  sample1, sample2, active);
+        active &= any(neq(depolarize(weight), 0.f));
```
wjakob, line 115: A minor comment here:  the `none(active)` call below will trigger an expensive horizontal operation to the GPU that flushes all computation and waits for an answer whether any of the entries is still alive. (So a kernel compilation, flush, and PCI express read that stalls for the GPU).

Because the answer is almost inevitably: -- yes, there is still one lane active --, we generally use `none_or<true>` which will use the default template argument on in the JIT-compiled GPU mode and perform the horizontal operation on all other targets.

In this case, you are only saving a few multiplications, so it's probably not worth it to have this extra check at all. I would just remove it entirely.

src/bsdfs/bumpmap.cpp
```diff
@@ -0,0 +1,210 @@
...
+        // Evaluate nested BSDF with perturbed shading frame
+        SurfaceInteraction3f perturbed_si(si);
+        perturbed_si.sh_frame = frame(si, active);
+        perturbed_si.wi       = perturbed_si.to_local(si.wi);
+        Vector3f perturbed_wo = perturbed_si.to_local(wo);
+
+        active &= Frame3f::cos_theta(wo) *
+                  Frame3f::cos_theta(perturbed_wo) > 0.f;
+
+        return m_nested_bsdf->eval(ctx, perturbed_si, perturbed_wo, active);
```
wjakob, line 141: I think that we at some point need some better documentation on masking (my fault). The policy is that a function is allowed to return any garbage when the mask flag is `false`. (@Speierers, we may want to revisit this policy at some point because it introduces inefficiencies).  For now, you will want to wrap this in another `select(active, ..., 0)`.

src/bsdfs/bumpmap.cpp
```diff
@@ -0,0 +1,210 @@
...
+        // Evaluate nested BSDF pdf with perturbed shading frame
+        SurfaceInteraction3f perturbed_si(si);
+        perturbed_si.sh_frame = frame(si, active);
+        perturbed_si.wi       = perturbed_si.to_local(si.wi);
+        Vector3f perturbed_wo = perturbed_si.to_local(wo);
+
+        active &= Frame3f::cos_theta(wo) *
+                  Frame3f::cos_theta(perturbed_wo) > 0.f;
+
+        return m_nested_bsdf->pdf(ctx, perturbed_si, perturbed_wo, active);
```
wjakob, line 157: ditto.

ommaury:
Hi Wenzel,
Addressed your comments. The same tweaks prob apply to the `normalmap` plugin I think.
Thanks,
Olivier

Speierers:
Changes in the `mitubsa-data` are good to go 🚀

Speierers:
Thanks @ommaury for this! I am going to merge this (manually as the commit history isn't very linear on your branch). Thanksfully there isn't to many files to change ;)

Speierers:
Done :rocket:

PR #213, opened 2020-07-15
Update doc to remove shape support on GPU

arpit15:
Removes warning from primitive shapes as they are now supported using Optix


Speierers:
Thanks for this

PR #243, opened 2020-08-10
Support for texture packing  [✨ new feature]

nmichiels:
This PR adds support for texture packing. This allows the user to select/filter one or multiple channels which are converted internally to a new monochromatic or rgb bitmap. The channels to select be chosen using the new 'channel' property of bitmap textures (as inspired by Mitsuba 0.6).

## Description
This required changes in three files:
- *\src\libcore\bitmap.cpp*: Added a function **extractChannels** to extract one or more channels out of an existing bitmap (ported code from Mitsuba 0.6). This requires a helper function **getBytesPerComponent** to idenitfy the number of bytes per pixel.
- *\include\mitsuba\core\bitmap.h Added function declarations for **extractChannels** and **getBytesPerComponent**
- *\src\textures\bitmap.cpp*: Code for parsing the channel property, which is a string, into a vector of channels to extract. Here the bitmap is altered and interchanged with the newly generated bitmap. When no *channel* property is provided, the bitmap stays unchanged.

## Testing

The PR is tested for rgb variant. Not tested yet for `gpu_*` and `packet_*` variants.

## Checklist:

- [ ] Testing for `gpu_*` and `packet_*` variants. 
- [ ] Documentation

### DIFF
--- a/include/mitsuba/core/bitmap.h
+++ b/include/mitsuba/core/bitmap.h
@@ -239,9 +239,8 @@ public:
     /// Move constructor
     Bitmap(Bitmap &&bitmap);
 
-
+    // Returns number of bytes per pixel channel
     int get_bytes_per_component() const;
-
 
     /// Return the pixel format of this bitmap
     PixelFormat pixel_format() const { return m_pixel_format; }
--- a/include/mitsuba/core/bitmap.h
+++ b/include/mitsuba/core/bitmap.h
@@ -239,6 +239,10 @@ public:
     /// Move constructor
     Bitmap(Bitmap &&bitmap);
 
+
+    int getBytesPerComponent() const;
+
+
     /// Return the pixel format of this bitmap
     PixelFormat pixel_format() const { return m_pixel_format; }
 
@@ -379,6 +383,13 @@ public:
     /// Equivalent to \ref write(), but executes asynchronously on a different thread
     void write_async(const fs::path &path, FileFormat format = FileFormat::Auto,
                      int quality = -1) const;
+
+    /**
+     * \brief Extract several color channels of a multi-channel
+     * bitmap and return them as a bitmap with the given pixel format
+     */
+    ref<Bitmap> Bitmap::extractChannels(PixelFormat fmt, const std::vector<int> &channels) const;
+
 
     /**
      * \brief Up- or down-sample this image to a different resolution
--- a/include/mitsuba/core/bitmap.h
+++ b/include/mitsuba/core/bitmap.h
@@ -240,7 +240,7 @@ public:
     Bitmap(Bitmap &&bitmap);
 
 
-    int getBytesPerComponent() const;
+    int get_bytes_per_component() const;
 
 
     /// Return the pixel format of this bitmap
--- a/include/mitsuba/core/bitmap.h
+++ b/include/mitsuba/core/bitmap.h
@@ -388,7 +388,7 @@ public:
      * \brief Extract several color channels of a multi-channel
      * bitmap and return them as a bitmap with the given pixel format
      */
-    ref<Bitmap> Bitmap::extractChannels(PixelFormat fmt, const std::vector<int> &channels) const;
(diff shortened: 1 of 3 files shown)


include/mitsuba/core/bitmap.h
```diff
@@ -380,6 +384,13 @@ class MTS_EXPORT_CORE Bitmap : public Object {
     void write_async(const fs::path &path, FileFormat format = FileFormat::Auto,
                      int quality = -1) const;
 
+    /**
+     * \brief Extract several color channels of a multi-channel
+     * bitmap and return them as a bitmap with the given pixel format
+     */
+    ref<Bitmap> Bitmap::extractChannels(PixelFormat fmt, const std::vector<int> &channels) const;
```
Speierers, line 391: Please use the `underscore_case` convention for the method / variable names:

```suggestion
    ref<Bitmap> Bitmap::extract_channels(PixelFormat fmt, const std::vector<int> &channels) const;
```

include/mitsuba/core/bitmap.h
```diff
@@ -239,6 +239,10 @@ class MTS_EXPORT_CORE Bitmap : public Object {
     /// Move constructor
     Bitmap(Bitmap &&bitmap);
 
+
+    int getBytesPerComponent() const;
```
Speierers, line 243: `underscore_case` convention
```suggestion
    int get_bytes_per_component() const;
```

include/mitsuba/core/bitmap.h
```diff
@@ -239,6 +239,10 @@ class MTS_EXPORT_CORE Bitmap : public Object {
...
+
+
```
Speierers, line 245: remove extra space 😉

src/textures/bitmap.cpp
```diff
@@ -205,6 +230,53 @@ class BitmapTexture final : public Texture<Float, Spectrum> {
         m_mean = ScalarFloat(mean / pixel_count);
     }
 
+    /**
+     * Parse the channel string into a vector of channels and a new pixel format based on the new number of channels
```
Speierers, line 234: Should we move this to `libcore/bitmap.h` instead?

src/textures/bitmap.cpp
```diff
@@ -205,6 +230,53 @@ class BitmapTexture final : public Texture<Float, Spectrum> {
...
+     * e.g.: r => [0] or rb => [0,2] or bg => [2,1] 
+     */
+    static std::vector<int> findChannels(const std::string channelStr, Bitmap::PixelFormat& new_fmt) {
```
Speierers, line 237:
```suggestion
    static std::vector<int> find_channels(const std::string channel_str, Bitmap::PixelFormat& new_fmt) {
```

src/textures/bitmap.cpp
nmichiels: I've added this as a pure helper function for parsing the channel string parameter in `src/textures/bitmap.cpp`. There is no declaration in the header file. I'm not sure if this should be a public reusable function in bitmap.h.

wjakob:
IMO this feature is a little too specialized to consider for the general Mitsuba 2 codebase. An IMO better way to support such transformations would be to improve the Python API so that users can directly provide a `Bitmap` instance to the `bitmap` plugin.

wjakob:
Hmm, I realize now that this was indeed a supported feature of Mitsuba 1 🤦 . I'm generally happy to accept such a feature, but it should be implemented much more succinctly. I would suggest an API
```
Bitmap *extract(const char &name);
```
instead of the current split into two functions in this PR. Furthermore, this function is easily realizable using `StructConverter` -- simply create an output struct that only has the desired channel, with the same component format.

wjakob:
Also note that Mitsuba 2 uses `underscore_case` and not `camelCase`.

Speierers:
I believe this is now supported in the `next` branch with the `Bitmap.split()` method. Therefore I will close this PR.

Issue #211, opened 2020-07-14
Spectral rendering


SergeyShlyaev:
Hello,

In case you'd like to keep track on how Mitsuba is used online. This isn't an issue but rather the opposite.
Wanted to thank you for making such interesting and feature-rich renderer.

Made a short video on spectral rendering utilizing some of awesome Mitsuba features:
![stopframe_halfres](https://user-images.githubusercontent.com/63272131/87369169-d67ea880-c54d-11ea-8be0-e9031c3c3577.jpg)
https://vimeo.com/437885264

Wondering if you'd be interested to hear about some of badly missed features to see if they align with your development plan?

Speierers:
Hi @SergeyShlyaev ,

Great work! Thanks for sharing this video, some of those clips look really nice!

We would be happy to hear about user requests and badly missed features. Please go ahead 😉

SergeyShlyaev:
Hi @Speierers,

Sure, here is the wishlist. It is mostly about getting data into Mitsuba.

#### Curves support
Smooth curves support, with option to render curves as cylinders or oriented ribbons. 
To render hair, fur, ropes, strings, grass.
Usually if there is “normal” attribute on curve, it is rendered as ribbon oriented according to normals, so we
can make twisted ribbons. If there is no “normal” it’s rendered as cylinder.
Would be super cool to have uv parametrization for curve, so u runs across the curve, and v is 0 at root and 1.0 at tip of the hair. So we can modify width along ‘v’, to achieve tapered cylinders or ribbons, like grass blades.

#### Alembic support
Alembics to transfer cameras and assets. https://www.alembic.io/ 
With support for arbitrary face and vertex attributes.
It’s kind of standard format for interchange now. Would be great to have option to read arbitrary face and vertex attributes for rendering. So we can write out balloons color or transmittance and use it in bsdf.

#### Usd support 
This is Universal scene description.
https://graphics.pixar.com/usd/docs/index.html
While alembics are for individual assets, usd is to transfer entire complex scenes. It’s also considered a standard now.
Project is open source: https://github.com/PixarAnimationStudios/USD
Apart from many other use cases, It is a base for Nvidia’s Omniverse platform.
https://developer.nvidia.com/nvidia-omniverse-platform

#### Subdiv support
Subdiv support, so poly meshes can be subdivided on the fly during rendering. 
Without having to output massive, high poly `ply` or `obj` files.
There is an opensource project from Pixar on it:
[...]

PR #260, opened 2020-08-21
Support for alembic meshes (www.alembic.io) - [✨ new feature]

SergeyShlyaev:
## Description
Alembic is an open computer graphics interchange framework. (www.alembic.io)
It is file format for exchanging 3D computer animation information between different platforms and applications. Any app can write their 3D information out in the Alembic format. Which can then be read in by any other app that supports Alembic, such as Maya, 3dsmax, Houdini, Katana, Nuke to name a few. 

Alembic stores 3D geometry in a format that is very efficient with a small file size that is quick to read and unpack. 
Using alembics makes it a ton easier to get complex scenes into Mitsuba.
Current plugin implementation supports arbitrary meshes made of triangles, quads or any n-gons with optional UV coordinates, vertex normals and other custom vertex or face attributes.

Mentioned in #211.

## Testing

Tried with various types of geometry, including n-gons, degenerate geo, various vertex and face attributes. Tested on light, medium and very heavy geo. Test scene attached.

## Questions
1.  How do I create multiple mesh shapes? Alembic may contain several separate meshes like serialized files, but now it only seem to create the last mesh in file or the one that is specified by shape_index. How can I create more then one? 
2. How can one assign material from inside cpp code?
We can specify material assignment in alembic file, would be great to be able to use it in plugin.
This may open the road to tighter integration cause we can specify bsdf assignment directly in DCC.
Supposedly, I’ll need to manipulate m_bsdf somehow? 
3. Is there a concept of faceset/vertex set in Mitsuba?
Faceset is a set (or group) of faces in an existing mesh. This is useful in order to more easily re-select them later when making shader, attribute, and visibility assignments to a sub-set of faces on a single mesh.

## Prerequisities 
Requires alembic
`git clone http://github.com/alembic/alembic`

## Checklist:
- [+] My code follows the [style guidelines](https://mitsuba2.readthedocs.io/en/latest/src/developer_guide/intro.html#introduction) of this project
- [+] My changes generate no new warnings
- [*] My code also compiles for `gpu_*` and `packet_*` variants. If you can't test this, please leave below.
Packet variants – Yes. Gpu can't test as I'm on Mac.

- [+] I have commented my code
Need help on how to do that. Doc for the plugin is included in cpp file.
Need help on how to do that
- [+] I cleaned the commit history and removed any "Merge" commits
- [+] I give permission that the Mitsuba 2 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba2/blob/master/LICENSE)

abc01
abc02

[mitsuba_alembic_test.zip](https://github.com/SergeyShlyaev/mitsuba2/files/5106648/mitsuba_alembic_test.zip)

### DIFF
--- a/src/shapes/CMakeLists.txt
+++ b/src/shapes/CMakeLists.txt
@@ -4,6 +4,7 @@ add_plugin(obj         obj.cpp)
 add_plugin(ply         ply.cpp)
 add_plugin(blender     blender.cpp)
 add_plugin(serialized  serialized.cpp)
+add_plugin(abc         abc.cpp)
 
 add_plugin(cylinder    cylinder.cpp)
 add_plugin(disk        disk.cpp)
--- a/src/shapes/abc.cpp
+++ b/src/shapes/abc.cpp
@@ -1,0 +1,608 @@
+#include <mitsuba/render/mesh.h>
+#include <mitsuba/core/fstream.h>
+#include <mitsuba/core/mstream.h>
+#include <mitsuba/core/fresolver.h>
+#include <mitsuba/core/properties.h>
+#include <mitsuba/core/util.h>
+#include <mitsuba/core/timer.h>
+#include <enoki/half.h>
+#include <unordered_map>
+#include <fstream>
+
+// Alembic Includes
+#include <Alembic/Abc/All.h>
+#include <Alembic/AbcGeom/All.h>
+#include <Alembic/AbcCoreFactory/All.h>
+#include <Alembic/AbcCoreOgawa/All.h>
+
+NAMESPACE_BEGIN(mitsuba)
+
+/**!
+
+.. _shape-abc:
+
+ABC (Alembic) mesh loader (:monosp:`abc`)
+----------------------------------------------------------
+
+.. pluginparameters::
+
+ * - filename
+   - |string|
+   - Filename of the ABC file that should be loaded 
+ * - shape_index
+   - |int|
+   - A :monosp:`.serialized` file may contain several separate meshes. This parameter
+     specifies which one should be loaded. (Default: 0, i.e. the first one)
+ * - m_flip_normals
+   - |bool|
+   - Will flip normals on the mesh. Some programs output face indexes in clockwise order, some in counter-clockwise. 
+     This parameter optionally fixes handedness. (Default: |true|)
+ * - load_visible
+   - |bool|
+   - Alembic allows for objects or specific faces of an object to be flagged as either visible or hidden. 
+     This is useful because it allows objects or faces to appear or disaper over time, by animating the visiblity property. 
+     Setting this to |true| will load only objects flagged as visible. (Default: |false|)
+ * - face_normals
+   - |bool|
+   - When set to |true|, any existing or computed vertex normals are
+     discarded and *face normals* will instead be used during rendering.
+     This gives the rendered object a faceted appearance. (Default: |false|)
+ * - flip_tex_coords
+   - |bool|
+   - Treat the vertical component of the texture as inverted? (Default: |true|)
+ * - to_world
+   - |Transform4f|
+   - Specifies an optional linear object-to-world Transform4fation.
+     (Default: none, i.e. object space = world space)
+
+This plugin implements loader for Alembic file format (www.alembic.io). The
+current plugin implementation supports arbitrary meshes with optional UV
+coordinates, vertex normals and other custom vertex or face attributes.
+
+ */
+
+
+using namespace std;
+using namespace Alembic::AbcGeom;
+using namespace Alembic::AbcCoreFactory;
+using namespace Alembic::AbcCoreAbstract;
+
+template <typename Float, typename Spectrum>
+class AlembicMesh final : public Mesh<Float, Spectrum> {
+public:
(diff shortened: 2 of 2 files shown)


Speierers:
Hi @SergeyShlyaev,

Thanks a lot for this PR!

> 1. How do I create multiple mesh shapes? Alembic may contain several separate meshes like serialized files, but now it only seem to create the last mesh in file or the one that is specified by shape_index. How can I create more then one?

For this you will need to use the `mitsuba::Object::expand()` method (see doc in `object.h`)

> 2. How can one assign material from inside cpp code?

You will need to instanciate the BSDF yourself using the `PluginManager` and assign the returned pointer to `Shape::m_bsdf`. 

> 3. Is there a concept of faceset/vertex set in Mitsuba?

No there isn't. And we probably won't support this in the future. Mitsuba 2 being a research-oriented framework, it is unlikely that we will add support for "fancy" shaders to do things like "shader, attribute, and visibility assignments".

I am worried about the `alembic` dependency. is this library self-contained? Or will it pull a massive amount of code from its own dependencies as well? We would rather keep Mitsuba as lightweight as possible ;)

wjakob:
Given the very heavy nature of Alembic, USD, and similar packages, I don't think that such a contribution can be merged  directly into Mitsuba master. :(. (The policy of Mitsuba 2 is that all dependencies are automatically built using a self-contained build system)

SergeyShlyaev:
Hi guys, 

Thank you for the tips.
There are only two required dependencies.
From https://github.com/alembic/alembic

1. Required:

>     CMake (2.8.11+ newer is better for Windows builds) www.cmake.org
>     OpenEXR (2.2.0) www.openexr.com (for ilmbase)

Both of them are already part of Mitsuba I believe, so Alembic does not seem to bring any new dependencies.

While working on spectral video, found myself spending too much time preparing datasets for render and debugging why something does not work. I believe there should be simpler way to do it. 
Alembic looks like natural choice for it due to vast format capabilities and wide adoption.

Having it should reduce preparation time. And free time for actual research work in Mitsuba which is amazing.

2. Multiple meshes
> For this you will need to use the mitsuba::Object::expand() method

Thanks, `expand()` works, it creates several meshes. But they do not seem to pick up material assignment from xml file or other properties like vertex normals.

Doing it like:
```
std::vector<ref<Object>> m_mesh_objects;
for (every polymesh)
{
     // read data from file
    // fill data arrays
    // create new mesh:
    Properties props;
    ref<Mesh<Float, Spectrum>>  mesh =
        new Mesh<Float, Spectrum>(m_mesh_name, m_vertex_count, m_face_count, 
                props, polymesh_has_vertex_normals, polymesh_has_uvs);
    mesh->parameters_changed();
    m_mesh_objects.push_back(mesh);
}
...
std::vector<ref<Object>> expand() const override {
         return m_mesh_objects;
}
```
What am I doing wrong? Need to make it pick up material specified in xml file as usual:
```
    <shape type="abc">
        <string name="filename" value="mesh/abc/test.abc"/> 
	<ref id="my_diffuse_bsdf"/>
    </shape>
```
Tech question: Do you recommend inherit AlembicMesh class from `Mesh` or from `Shape`?

3. Facesets
> Mitsuba 2 being a research-oriented framework, it is unlikely that we will add support for "fancy" shaders to do things like "shader, attribute, and visibility assignments”.

Sure. Actually, Blender addon has very similar feature for exporting mesh with multiple materials assigned. For Alembics we could do it same way if needed, it answers the question. 
Visibility assignments are already supported, no need to do anything.

Cheers

SergeyShlyaev:
> You will need to instanciate the BSDF yourself using the PluginManager and assign the returned pointer to Shape::m_bsdf

Thank you, found the example in shape.cpp:80 on how to create bsdf. But is there a way to specify bsdf by name, not to create bsdf? 
Similar to how it's done in xml:
```xml
<ref id="my_diffuse_bsdf"/>
```
This is for extra option to be able to follow material assignment from Alembic file.
To make it probably I need to construct named reference somehow and use it in `props`?

Speierers:
> But is there a way to specify bsdf by name, not to create bsdf?

Would this name correspond to the `id` of a BSDF defined elsewhere in the XML file? 

I suggest we focus on the "geometry" part of this plugin for now. Adding support for BSDF binding might require changes in the rest of the codebase which would make this PR quite heavy.

SergeyShlyaev:
Pushed new version with support for loading multiple objects from file.
Added support for `shape_name` to specify object by name.

alembic_sweets

> Would this name correspond to the id of a BSDF defined elsewhere in the XML file?

Yes, correct.

> I suggest we focus on the "geometry" part of this plugin for now. Adding support for BSDF binding might require changes...

Agree, it looks slightly more complicated after deeper look. Probably to make it work, need to instantiate bsdf and all of it’s nested references manually and attach it via `props`. Similar to `instantiate_node()` function in xml.cpp

[alembic_sweets.zip](https://github.com/mitsuba-renderer/mitsuba2/files/5153593/alembic_sweets.zip)

SergeyShlyaev:
Added camera plugin to read camera data fom alembic file.
Adding this fullfills the original intention of this PR, to have simpler way of getting complex datasets into renderer.
With this one can export scene with geo, lights, cameras into single Alembic file and then read it into Mitsuba.

city

PR #281, opened 2020-09-21
[Feature] Python bindings for Volume API

leroyvn:
This small PR adds missing Python bindings for the volume API.

## Description

This small PR adds missing Python bindings for the volume API. They are required to write tests for new `Volume` plugins. I'll be happy to fix issues if, but I may need some guidance given my lack of experience with bindings writing.

## Testing

⚠️ _I did not include the tests I wrote which helped me check that these changes work: I used these bindings to write tests for a new plugin not suitable for a PR yet._

## Checklist:

- [ ] My code also compiles for `gpu_*` and `packet_*` variants (_only checked for `packet_*` variants_)
- [x] ~I have commented my code~ (_not applicable_)
- [ ] I have made corresponding changes to the documentation (_I don't know if any action is required here_)
- [ ] I have added tests that prove my fix is effective or that my feature works (_see comment above_)
- [x] I cleaned the commit history and removed any "Merge" commits


src/librender/python/texture3d_v.cpp
```diff
@@ -1,33 +1,29 @@
-#include <mitsuba/core/properties.h>
 #include <mitsuba/render/interaction.h>
-#include <mitsuba/render/spectrum.h>
+#include <mitsuba/render/texture.h>
 #include <mitsuba/python/python.h>
 
 MTS_PY_EXPORT(Volume) {
+    MTS_PY_IMPORT_TYPES(Volume)
     MTS_PY_CLASS(Volume, Object)
-        .def(py::init<const Properties &>())
```
Speierers, line 8: We should be able to construct a `Volume` using the `Properties` constructor as well. Did you remove this line for any specific reason?

src/librender/python/texture3d_v.cpp
```diff
@@ -1,33 +1,29 @@
-#include <mitsuba/core/properties.h>
 #include <mitsuba/render/interaction.h>
-#include <mitsuba/render/spectrum.h>
+#include <mitsuba/render/texture.h>
 #include <mitsuba/python/python.h>
 
 MTS_PY_EXPORT(Volume) {
```
Speierers, line 5: Let's move all of this this to the `texture_v.cpp` file.

Speierers:
Hi @leroyvn ,

Thanks a lot for this PR. This is looking pretty good, just a few minor comments.

src/librender/python/texture3d_v.cpp
leroyvn: Not at all, I just missed it. I'll fix that.

leroyvn: Now that I've given it a try, I start seeing the reason why it's not done. I basically copied binding code from those for `Texture`, why also don't have a constructor from `Properties`. I tried simply adding the usual `.def(py::init<const Properties &>())` line to the bindings and got an error like this:
```
In file included from ../src/librender/python/texture_v.cpp:4:
In file included from ../include/mitsuba/python/python.h:10:
In file included from ../ext/pybind11/include/pybind11/pybind11.h:47:
../ext/pybind11/include/pybind11/detail/init.h:63:68: error: calling a protected constructor of class 'mitsuba::Volume<float, Color<float, 3> >'
```
The problem is that the `Volume` (and `Texture`) constructors from `Properties` are protected. From what I understand, other abstract classes with protected constructors either don't have bindings for their constructor from `Properties` (_e.g._ `Endpoint`) or have a "proxy" class with a public constructor (_e.g._ `Sensor`).

I added this intermediate class for `Volume`. I made it very minimal, I hope it's enough.

src/librender/python/texture3d_v.cpp
leroyvn: Done!

src/librender/python/texture3d_v.cpp
Speierers: Actually this makes sense, my bad. Let's not expose this constructor for now. We can always instanciate a `Volume` plugin using the `xml.load_dict()` or `xml.load_string()` methods.

In the future, we will add support for custom python `Volume` plugin, similar to what is done with `BSDF`, `Integrator` ,...

leroyvn: Ok, I rolled back this change. I also squashed the commits.

Speierers:
Thanks @leroyvn for your great work! :rocket:

PR #304, opened 2020-10-26
[feature] Light tracer integrator ☀️→ 🎥 

merlinND:
## Description

Adds a particle tracer (= light tracer), and the required new interfaces.

- [x] Rebase onto `next`
- [x] Interface upgrades
- [x] New scene methods
- [x] New endpoint methods
- [x] `ParticleTracer` integrator
- [ ] Discuss and agree on added interfaces
- [ ] Support all emitters
  - [x] Area
  - [ ] Area (textured)
  - [x] Constant
  - [x] Envmap
  - [x] Directional
  - [ ] Point
  - [ ] Projector
  - [ ] Spot
- [ ] Figure out cache misses when repeatedly calling the integrator during optimization
- [ ] Support all modes (should at least compile)
- [x] <strike>Support ray differentials</strike>  (will not be supported)
- [x] Add Python bindings
- [ ] Check for additional leftover TODOs in the code
- [ ] Add documentation

- [ ] Backport `serialized` fix to master (122b636)
- [ ] Backport `path` fix to master (78b65d6)

## Testing

- [ ] Add unit tests
- [ ] Add render tests
  - [ ] Create scene(s) containing all types emitters
  - [ ] Should match path tracer at all `max_depth` values (0..8)
- [ ] Edge cases to test specifically:
  - [ ] Scene bbox smaller than emitter + infinite emitter
  - [ ] `sample_visible_emitters` + sensor that is not a single point + infinite emitter

## Checklist

- [ ] My code follows the [style guidelines](https://mitsuba2.readthedocs.io/en/latest/src/developer_guide/intro.html#introduction) of this project
- [ ] My changes generate no new warnings
- [ ] My code also compiles for `gpu_*` and `packet_*` variants.
- [ ] I have commented my code


merlinND:
Hi @wjakob and @Speierers,

This is an initial version of the particle tracer integrator. It can render the Cornell box in `scalar_rgb`, but still needs some work for infinite emitters.

Before implementing them in all plugins, could you please take a look at the new methods of `Endpoint` and `Scene`?

include/mitsuba/render/integrator.h
```diff
@@ -222,6 +224,91 @@ class MTS_EXPORT_RENDER MonteCarloIntegrator : public SamplingIntegrator<Float,
...
+    MTS_IMPORT_BASE(Integrator, should_stop, m_stop, m_timeout, m_render_timer)
+    MTS_IMPORT_TYPES(Scene, Sensor, Film, BSDF, ImageBlock, Sampler, EmitterPtr)
+    using RadianceSample3f = int; // TODO: this type no longer exists
+
+    explicit LightTracerIntegrator(const Properties &props);
+    virtual ~LightTracerIntegrator();
+
+    /// Perform the main rendering job
+    bool render(Scene *scene, Sensor *sensor) override;
+    void cancel() override;
```
Speierers, line 244: Should this method be moved to the `Integrator` interface?

src/librender/imageblock.cpp
```diff
@@ -168,6 +168,25 @@ ImageBlock<Float, Spectrum>::put(const Point2f &pos_, const Float *value, Mask a
     return active;
 }
 
+MTS_VARIANT void
+ImageBlock<Float, Spectrum>::overwrite_channel(size_t channel,
+                                               ScalarFloat value) {
+    size_t pixel_count = (width() + m_border_size) * (height() + m_border_size);
+    Assert(m_data.size() == m_channel_count * pixel_count);
+
+    // TODO: any way to avoid this specialization?
```
Speierers, line 177: `m_data` being an `DynamicBuffer<Float>` you should be able to use `ek::scatter` for the `scalar_*` variants as well. Maybe something like
```c++
auto indices = m_channel_count * ek::arange<DynamicBuffer<UInt64>>(pixel_count) + channel;
ek::scatter(m_data, value, indices);
```

src/librender/integrator.cpp
```diff
@@ -312,11 +319,190 @@ MTS_VARIANT MonteCarloIntegrator<Float, Spectrum>::MonteCarloIntegrator(const Pr
...
+            for (auto i = range.begin(); i != range.end() && !should_stop(); ++i) {
+                if (likely(m_max_depth != 0)) {
+                    // Account for visible emitters
+                    sample_visible_emitters(scene, sensor, sampler, block);
+                }
+
+                // Primary & further bounces illumination
+                auto [ray, throughput] = prepare_ray(scene, sensor, sampler);
+
+                // TODO(!): how to create a correct initial si?
```
Speierers, line 412: For scalar variants this initialization should be okay. Otherwise we usually use `ek::zero<SurfaceInteraction3f>()` and then set `si.t = ek::Infinity<Float>` to ensure the surface interaction is invalid.

src/librender/scene.cpp
```diff
@@ -172,43 +172,117 @@ Scene<Float, Spectrum>::ray_test(const Ray3f &ray, Mask active) const {
...
+                       const typename Scene<Float, Spectrum>::EmitterPtr>
+Scene<Float, Spectrum>::sample_emitter_ray(Float time, Float sample1,
+                                           const Point2f &sample2,
+                                           const Point2f &sample3,
+                                           Mask active) const {
+    MTS_MASKED_FUNCTION(ProfilerPhase::SampleEmitterRay, active);
+    using EmitterPtr = ek::replace_scalar_t<Float, Emitter*>;
+
+    if (unlikely(m_emitters.empty()))
+        return std::make_tuple(Ray3f(), Spectrum(0.f), EmitterPtr(nullptr));
```
Speierers, line 185:
```suggestion
        return { Ray3f(), Spectrum(0.f), EmitterPtr(nullptr) };
```

src/librender/scene.cpp
```diff
@@ -172,43 +172,117 @@ Scene<Float, Spectrum>::ray_test(const Ray3f &ray, Mask active) const {
...
+    auto [ray, ray_weight] =
+        emitter->sample_ray(time, wavelengths, sample2, sample3, active);
+    // Account for the discrete probability of sampling this emitter.
+    ray_weight *= wav_weight * emitter_weight;
+#else
+    auto [ray, ray_weight] =
+        emitter->sample_ray(time, sample1, sample2, sample3, active);
+#endif
+
+    return std::make_tuple(ray, ray_weight, emitter);
```
Speierers, line 206:
```suggestion
    return { ray, ray_weight, emitter };
```

include/mitsuba/render/scene.h
```diff
@@ -79,6 +79,73 @@ class MTS_EXPORT_RENDER Scene : public Object {
...
+     *
+     * \param sample
+     *    A uniformly distributed number in [0, 1).
+     *
+     * \return
+     *    The index of the chosen emitter along with the sampling weight (equal
+     *    to the inverse PDF), and the rescaled random sample.
+     */
+    std::tuple<UInt32, Float, Float>
+    sample_emitter_reuse(Float index_sample, Mask active = true) const;
```
Speierers, line 108: Could this method take the random number as reference and update it "in-place"? In which case we wouldn't both methods actually. Or is there a case where we explicitly want both?

merlinND:
@wjakob Let me know if you'd like me to rebase onto `next` before review (I don't know how big of a gap there will be).

.gitmodules
```diff
@@ -36,7 +36,7 @@
     url = https://github.com/mitsuba-renderer/embree
 [submodule "ext/enoki"]
 	path = ext/enoki
-	url = https://github.com/wjakob/enoki
+	url = git@github.com:wjakob/enoki.git
```
wjakob, line 39: Let's undo this, we should definitely clone via `https://` instead of the old-school git protocol. or am I misunderstanding this line?

include/mitsuba/render/endpoint.h
```diff
@@ -115,6 +115,52 @@ class MTS_EXPORT_RENDER Endpoint : public Object {
                      const Point2f &sample,
                      Mask active = true) const;
 
+    /**
+     * \brief Independently of the scene, sample a position on the endpoint.
```
wjakob, line 119: This ("independent of the scene", "independent of transport phenomena") sounds a little strange to me. Mitsuba 1 had documentation for all of these methods -- can you copy-paste from there as appropriate?

include/mitsuba/render/endpoint.h
```diff
@@ -115,6 +115,52 @@ class MTS_EXPORT_RENDER Endpoint : public Object {
...
+     */
+    virtual std::pair<PositionSample3f, Float>
+    sample_position(Float time, const Point2f &sample,
+                    Mask active = true) const;
+
+    /**
+     * \brief Importance sample a set of wavelengths proportional to the
+     * sensitivity/emission spectrum.
+     * Any discrepancies between ideal and actual sampled profile are absorbed
+     * into a spectral importance weight that is returned along with the ray.
```
wjakob, line 146: Add comment to clarify what happens here in RGB / mono mode

src/integrators/ptracer.cpp
```diff
@@ -0,0 +1,245 @@
...
+#endif
+
+NAMESPACE_BEGIN(mitsuba)
+
+/**
+ * Traces rays from the light source and attempts to connect them to the sensor
+ * at each bounce.
+ */
+template <typename Float, typename Spectrum>
+class ParticleTracerIntegrator : public LightTracerIntegrator<Float, Spectrum> {
```
wjakob, line 32: The particle tracer is a sufficiently special snowflake that I don't think we will have another integrator like it. Given that, I think that it would be better to have the specialization of the `render()` method directly here in the class.

src/integrators/ptracer.cpp
```diff
@@ -0,0 +1,245 @@
...
+    std::pair<Spectrum, Float>
+    trace_light_ray(Ray3f ray, const Scene *scene, const Sensor *sensor,
+                    Sampler *sampler, SurfaceInteraction3f &si,
+                    Spectrum throughput, int depth, ImageBlock *block,
+                    Mask active) const override {
+        // Tracks radiance scaling due to index of refraction changes
+        Float eta(1.f);
+        BSDFContext ctx(TransportMode::Importance);
+
+        // TODO: support ray differentials
```
wjakob, line 68: Remove TODO, we don't support ray differentials for this integrator.

src/integrators/ptracer.cpp
```diff
@@ -0,0 +1,245 @@
...
+        // Sample one ray from an emitter in the scene.
+        auto [ray, ray_weight, unused] = scene->sample_emitter_ray(
+            time, wavelength_sample, direction_sample, position_sample, active);
+        return std::make_pair(ray, ray_weight);
+    }
+
+    std::pair<Spectrum, Float>
+    trace_light_ray(Ray3f ray, const Scene *scene, const Sensor *sensor,
+                    Sampler *sampler, SurfaceInteraction3f &si,
+                    Spectrum throughput, int depth, ImageBlock *block,
```
wjakob, line 62: `depth` should likely not be a parameter (or it's at least not clear to me why it is.)

include/mitsuba/render/endpoint.h
```diff
@@ -115,6 +115,52 @@ class MTS_EXPORT_RENDER Endpoint : public Object {
                      const Point2f &sample,
                      Mask active = true) const;
 
+    /**
+     * \brief Independently of the scene, sample a position on the endpoint.
+     *
+     * This is entirely independent of transport phenomena, and samples
+     * uniformly over the surface area.
+     *
+     * For infinite area lights, the returned point is placed on a bounding
```
wjakob, line 124: Hrmmm, I must admit I am not a fan of this interface.

What I would prefer is the convention that envmap emitters only support the `sample_ray` interface that is actually more complex: it picks a direction and then an offset on a disk (which is the projection of the bounding sphere onto the sampled direction).

include/mitsuba/render/endpoint.h
```diff
@@ -115,6 +115,52 @@ class MTS_EXPORT_RENDER Endpoint : public Object {
...
+     *     A \ref PositionSample instance describing the generated sample
+     *     along with a spectral importance weight.
+     */
+    virtual std::pair<PositionSample3f, Float>
+    sample_position(Float time, const Point2f &sample,
+                    Mask active = true) const;
+
+    /**
+     * \brief Importance sample a set of wavelengths proportional to the
+     * sensitivity/emission spectrum.
```
wjakob, line 144: You need a newline here, otherwise the following text gets merged into the `\brief` part.

include/mitsuba/render/endpoint.h
```diff
@@ -130,6 +176,29 @@ class MTS_EXPORT_RENDER Endpoint : public Object {
...
+     * \brief Evaluate the probability density of the position sampling
+     * method implemented by the \ref sample_position() method.
+     * In simple cases, this will equal (1 / area of the endpoint).
+     *
+     * \param ps
+     *    The sampled position record.
+     * \return
+     *    The corresponding sampling density (for each wavelength).
+     */
+    virtual Float pdf_position(const PositionSample3f &ps, Mask active = true) const;
```
wjakob, line 189: co-locate pdf() and sample() methods.

include/mitsuba/render/film.h
```diff
@@ -31,6 +31,9 @@ class MTS_EXPORT_RENDER Film : public Object {
     /// Develop the film and write the result to the previously specified filename
     virtual void develop() = 0;
 
+    /// Overwrite the weight channel to the given value
```
wjakob, line 34: Can we remove this interface? Mitsuba 1 did not need this to implement the particle tracer.

include/mitsuba/render/imageblock.h
```diff
@@ -140,6 +140,9 @@ class MTS_EXPORT_RENDER ImageBlock : public Object {
      */
     Mask put(const Point2f &pos, const Float *value, Mask active = true);
 
+    /// Overwrite entire channel with the given value.
+    void overwrite_channel(size_t channel, ScalarFloat value);
```
wjakob, line 144: Can we remove this interface? Mitsuba 1 did not need this to implement the particle tracer.

include/mitsuba/render/integrator.h
```diff
@@ -222,6 +224,91 @@ class MTS_EXPORT_RENDER MonteCarloIntegrator : public SamplingIntegrator<Float,
...
+/**
+ * Base class for integrators that start paths from the light source, as opposed
+ * to sensor-based integrators.
+ */
+template <typename Float, typename Spectrum>
+class MTS_EXPORT_RENDER LightTracerIntegrator : public Integrator<Float, Spectrum> {
+public:
+    MTS_IMPORT_BASE(Integrator, should_stop, m_stop, m_timeout, m_render_timer)
+    MTS_IMPORT_TYPES(Scene, Sensor, Film, BSDF, ImageBlock, Sampler, EmitterPtr)
+    using RadianceSample3f = int; // TODO: this type no longer exists
```
wjakob, line 237: Sorry :-)

include/mitsuba/render/scene.h
```diff
@@ -79,6 +79,73 @@ class MTS_EXPORT_RENDER Scene : public Object {
...
+     *
+     * \return (ray, importance weight, emitter, radiance)
+     *    ray: sampled ray, starting from the surface of an emitter in the scene
+     *    importance weight: accounts for the difference between the profile and
+     *        the sampling density function actualy used.
+     *    emitter: pointer to the selected emitter.
+     *    radiance: emitted radiance along the sampled ray.
+     */
+    std::tuple<Ray3f, Spectrum, const EmitterPtr>
+    sample_emitter_ray(Float time, Float sample1, const Point2f &sample2,
```
wjakob, line 146: great!

src/emitters/constant.cpp
```diff
@@ -62,20 +65,29 @@ class ConstantBackgroundEmitter final : public Emitter<Float, Spectrum> {
...
         Vector3f v1 = warp::square_to_cosine_hemisphere(sample3);
-
-        return std::make_pair(Ray3f(m_bsphere.center + v0 * m_bsphere.radius,
-                                    Frame3f(-v0).to_world(v1), time, wavelengths),
-                              unpolarized<Spectrum>(weight) * (4.f * ek::sqr(ek::Pi<Float> * m_bsphere.radius)));
+        Vector3f direction = Frame3f(-v0).to_world(v1);
+
+        // 3. Sample spectrum
+        // TODO: how to best construct this `si`?
+        SurfaceInteraction3f si;
```
wjakob, line 78: You will probably have to zero-initialize this, or the GPU version will throw an exception in the virtual function call.
Actually, I am wondering if we can perhaps have `sample_wavelengths` *not* depend on the specific position?

src/emitters/envmap.cpp
```diff
@@ -130,8 +131,12 @@ class EnvironmentMapEmitter final : public Emitter<Float, Spectrum> {
 
     void set_scene(const Scene *scene) override {
         m_bsphere = scene->bbox().bounding_sphere();
-        m_bsphere.radius = ek::max(math::RayEpsilon<Float>,
-                               m_bsphere.radius * (1.f + math::RayEpsilon<Float>));
+        // m_bsphere.radius = ek::max(math::RayEpsilon<Float>,
+        //                        m_bsphere.radius * (1.f + math::RayEpsilon<Float>));
+        m_bsphere.radius = ek::max(math::RayEpsilon<Float>, m_bsphere.radius * 1.5f);
```
wjakob, line 136: hmm?

src/emitters/envmap.cpp
```diff
@@ -180,16 +226,55 @@ class EnvironmentMapEmitter final : public Emitter<Float, Spectrum> {
...
 
-        return {
-            ds,
-            unpolarized<Spectrum>(eval_spectrum(uv, it.wavelengths, active)) / ds.pdf
-        };
+        auto weight = unpolarized<Spectrum>(eval_spectrum(uv, it.wavelengths, active)) / ds.pdf;
+        return { ds, ek::select(ds.pdf > 0.f, weight, 0.f) };
+    }
+
+    /**
```
wjakob, line 242: TODO: remove

src/emitters/constant.cpp
```diff
@@ -105,13 +117,47 @@ class ConstantBackgroundEmitter final : public Emitter<Float, Spectrum> {
...
+    }
+
     Float pdf_direction(const Interaction3f &, const DirectionSample3f &ds,
                         Mask active) const override {
         MTS_MASKED_FUNCTION(ProfilerPhase::EndpointEvaluate, active);
 
         return warp::square_to_uniform_sphere_pdf(ds.d);
     }
 
+    Float pdf_position(const PositionSample3f & /*ps*/,
```
wjakob, line 155: TODO: remove

src/emitters/constant.cpp
```diff
@@ -105,13 +117,47 @@ class ConstantBackgroundEmitter final : public Emitter<Float, Spectrum> {
         );
     }
 
+    std::pair<PositionSample3f, Float>
```
wjakob, line 120: TODO: remove

src/integrators/ptracer.cpp
```diff
@@ -0,0 +1,245 @@
...
+        auto on_infinite_emitter = active && eq(si.shape, nullptr) && eq(bsdf, nullptr);
+        if (ek::any_or<true>(on_infinite_emitter)) {
+            auto right_side = Frame3f::cos_theta(local_d) > 0.f;
+            surface_weight[on_infinite_emitter] = select(right_side, surface_weight, 0.f);
+        }
+
+        result = weight * sensor_val * surface_weight;
+
+        // Splatting
+        if (block != nullptr) {
```
wjakob, line 222: What would be the use of calling this function without splatting enabled?

src/librender/scene.cpp
```diff
@@ -172,43 +172,117 @@ Scene<Float, Spectrum>::ray_test(const Ray3f &ray, Mask active) const {
...
+    MTS_MASKED_FUNCTION(ProfilerPhase::SampleEmitter, active);
+
+    size_t emitters_size = m_emitters.size();
+    // Fast paths if there is less than two emitters
+    if (unlikely(emitters_size == 0))
+        return { UInt32(-1), 0.f, index_sample };
+    if (emitters_size == 1)
+        return { 0, 1, index_sample };
+
+#if 0
```
wjakob, line 244: remove #if 0 part?

src/librender/scene.cpp
```diff
@@ -172,43 +172,117 @@ Scene<Float, Spectrum>::ray_test(const Ray3f &ray, Mask active) const {
...
+    ray_weight *= wav_weight * emitter_weight;
+#else
+    auto [ray, ray_weight] =
+        emitter->sample_ray(time, sample1, sample2, sample3, active);
+#endif
+
+    return std::make_tuple(ray, ray_weight, emitter);
+}
+
+MTS_VARIANT std::pair<typename Scene<Float, Spectrum>::UInt32, Float>
```
wjakob, line 209: Can we only have one of `sample_emitter` and `sample_emitter_reuse`? (I would just keep the `_reuse` variant and then remove that suffix from the name)

src/librender/scene.cpp
```diff
@@ -172,43 +172,117 @@ Scene<Float, Spectrum>::ray_test(const Ray3f &ray, Mask active) const {
...
     using EmitterPtr = ek::replace_scalar_t<Float, Emitter*>;
 
     Point2f sample(sample_);
     DirectionSample3f ds;
     Spectrum spec;
 
     size_t emitters_size = m_emitters.size();
-
     if (likely(emitters_size > 0)) {
         if (emitters_size == 1) {
```
wjakob, line 270: While you are changing this function, you can remove the `emitters_size == 1` special case. Enoki now performs the 1-instance-only optimization automatically internally.

src/librender/scene.cpp
```diff
@@ -172,43 +172,117 @@ Scene<Float, Spectrum>::ray_test(const Ray3f &ray, Mask active) const {
...
-
-            // Rescale sample.x() to lie in [0,1) again
-            sample.x() = (sample.x() - index*emitter_pdf) * emitters_size;
-
-            EmitterPtr emitter = ek::gather<EmitterPtr>(m_emitters_ptr.data(), index, active);
+            auto [index, emitter_weight, sample_re] =
+                sample_emitter_reuse(sample.x(), active);
+            sample.x() = sample_re;
+            // Account for the discrete probability of sampling this emitter
+            ds.pdf *= ek::rcp(emitter_weight);
```
wjakob, line 280: This line seems weird just from a convention point of view (Monte Carlo weights can contain all sorts of things that are not necessarily PDFs). Can we instead return the PDF and then multiply by it here? And then `rcp()` that for the weight.

src/librender/scene.cpp
```diff
@@ -172,43 +172,117 @@ Scene<Float, Spectrum>::ray_test(const Ray3f &ray, Mask active) const {
...
+            spec *= emitter_weight;
 
             // Sample a direction towards the emitter
+            EmitterPtr emitter = ek::gather<EmitterPtr>(m_emitters_ptr.data(), index, active);
```
wjakob, line 284: This has likely changed quite a bit in `next`.

src/librender/scene.cpp
```diff
@@ -227,6 +301,14 @@ Scene<Float, Spectrum>::sample_emitter_direction(const Interaction3f &ref, const
     return { ds, spec };
 }
 
+MTS_VARIANT Float
+Scene<Float, Spectrum>::pdf_emitter(UInt32 /*index*/, Mask /*active*/) const {
+    size_t emitters_size = m_emitters.size();
+    if (emitters_size == 0)
+        return 0.f;
+    return ek::rcp(emitters_size);
```
wjakob, line 309: It's not clear to me that `rcp` of an integer will give the right thing (`rcp` returns the same type IIRC, i.e. an integer!). This generally isn't an Enoki array, so using `ek::rcp` also feels strange. I would just write `1.f / m_emitters.size()`

wjakob (changes requested):
Hi Merlin -- this looks like a great start! I've added a few comments throughout. It looks to me like the next big steps are:

1. Rebase onto `next`.

2. Ensure that path tracer and particle tracer agree for different values of max_depth, and with different light source types (point, area, envmap). Naturally it generally won't match for specular things, so I'd only test smooth/microfacet materials.

3. LLVM, CUDA modes (megakernel via ek.Loop) -- Sébastien could probably help with those.

.gitmodules
merlinND: This was back when `wjakob/enoki` was private, and HTTPS cloning is not convenient when 2fac is enabled (`git@github` uses SSH instead).

wjakob: Ah, that makes a ton of sense :-)

include/mitsuba/render/film.h
merlinND: Mitsuba 1 used that API instead (pretty similar). Do you prefer the old one?

https://github.com/mitsuba-renderer/mitsuba/blob/cfeb7766e7a1513492451f35dc65b86409655a7b/include/mitsuba/render/film.h#L54-L55

include/mitsuba/render/endpoint.h
merlinND: That wold be nice if we can achieve it. For now this was used in `sample_visible_emitters` (emitters directly visible from the sensor). If we sample a ray from the emitter, there's zero probability it will connect with the sensor.

What I could do though is have a special case in the integrator: if it's an infinite emitter, only `sample_direction`, otherwise `sample_position`.

src/emitters/constant.cpp
merlinND: Since `sample_wavelengths` returns a sampling weight that includes the emitter's radiance, it needs a position for textured emitters I think?

src/librender/scene.cpp
merlinND: I agree that this line looks wrong because it silently assumes that `sample weight == 1 / pdf`. That being said, I would prefer keeping `sample_emitter`'s return value a sampling weight because (as far as I know), every other method starting with `sample_` returns a sampling weight.
Instead, how about: `ds.pdf = pdf_emitter(index);`?

merlinND: Also, I moved back these two lines (update of `ds.pdf` and `spec`) below the `std::tie()`, otherwise their effect was immediately ignored.

include/mitsuba/render/endpoint.h
merlinND: Update: managed to get rid of `sample_position` for infinite emitters by adding a special case in the integrator.

include/mitsuba/render/scene.h
merlinND: Kept only `sample_emitter_reuse` but renamed it `sample_emitter`, indeed we don't really need both.

merlinND:
Rebased onto `next`.

src/librender/scene.cpp
wjakob: Your suggestion sounds good.

src/librender/imageblock.cpp
merlinND: Thanks for the tip, I just had to add a cast because `value` was a scalar:

```
    auto indices =
        m_channel_count * ek::arange<DynamicBuffer<UInt64>>(pixel_count) + channel;
    ek::scatter(m_data, DynamicBuffer<Float>(value), indices);
```

docs/examples/10_inverse_rendering/invert_caustic.py
```diff
@@ -193,18 +198,88 @@ def load_ref_image(config, resolution, output_dir):
...
+    # to_world = ScalarTransform4f.rotate(axis=(1, 0, 0), angle=90)
+    # to_world = np.array(to_world.matrix)[:3, :3]
+    # vertices = (to_world @ vertices.T).T
+    assert vertices.shape == (n_vertices, 3)
+
+    texcoords =  np.stack([
+        U.ravel(), V.ravel()
+    ], axis=1)
+    assert texcoords.shape == (n_vertices, 2)
+
```
Speierers, line 233: Would it be possible to write this using `enoki` instead of `numpy`? It would be a great example on how to use the enoki Python API :)

merlinND:
Moved PR to `next` repo.

PR #307, opened 2020-10-27
[WIP] Suspending gradients for unbiased differentiable rendering

bathal1:
This PR allows to stop keeping track of derivatives while not recomputing intermediate results.

### DIFF
--- a/include/mitsuba/core/distr_1d.h
+++ b/include/mitsuba/core/distr_1d.h
@@ -233,6 +233,13 @@ public:
         };
     }
 
+    void set_grad_suspended(bool state) {
+        if constexpr (ek::is_diff_array_v<Float>) {
+            ek::set_grad_suspended(m_pmf, state);
+            ek::set_grad_suspended(m_cdf, state);
+        }
+    }
+
 private:
     FloatStorage m_pmf;
     FloatStorage m_cdf;
@@ -493,6 +500,13 @@ public:
 
         return { ek::fmadd(Float(index) + t, m_interval_size, m_range.x()),
                  ek::fmadd(t, y1 - y0, y0) * m_normalization };
+    }
+
+    void set_grad_suspended(bool state) {
+        if constexpr (ek::is_diff_array_v<Float>) {
+            ek::set_grad_suspended(m_pdf, state);
+            ek::set_grad_suspended(m_cdf, state);
+        }
     }
 
 private:
@@ -787,6 +801,14 @@ public:
 
         return { ek::fmadd(t, w, x0),
                  ek::fmadd(t, y1 - y0, y0) * m_normalization };
+    }
+
+    void set_grad_suspended(bool state) {
+        if constexpr (ek::is_diff_array_v<Float>) {
+            ek::set_grad_suspended(m_pdf, state);
+            ek::set_grad_suspended(m_cdf, state);
+            ek::set_grad_suspended(m_nodes, state);
+        }
     }
 
 private:
--- a/include/mitsuba/core/object.h
+++ b/include/mitsuba/core/object.h
@@ -115,6 +115,29 @@ public:
      */
     virtual std::string to_string() const;
 
+
+    /**
+     * \brief Suspend or resume tracking of derivatives.
+     *
+     * One important use case of Mitsuba 2 entails the propagation of
+     * derivatives through the full rendering process, typically via
+     * automatic differentiation. To this end, certain variants of the
+     * renderer perform all computation via Enoki's DiffArray<T> class, which
+     * internally builds a computation graph of all involved arithmetic.
+     * In certain situations, this is undesirable, and we need to temporarily
+     * suspend the propagation of gradients. That is typically done using the
+     * Object::traverse mechanism. However, some parameters may not be exposed
+     * through traverse, and we need to suspend them too, otherwise parts of
+     * the computation will still be recorded. This method handles such cases:
+     * its default implementation doed nothing, but classes using such "hidden"
+     * parameters may override it to properly suspend the tracking of their
+     * gradients.
+     *
+     * \param state
+     *      Whether to stop or resume recording derivatives.
+     */
+    virtual void set_grad_suspended(bool state);
+
 protected:
     /** \brief Virtual protected deconstructor.
      * (Will only be called by \ref ref)
--- a/src/libcore/object.cpp
+++ b/src/libcore/object.cpp
@@ -38,5 +38,7 @@ std::ostream& operator<<(std::ostream &os, const Object *object) {
     return os;
 }
 
+void Object::set_grad_suspended(bool /*state*/) { }
+
 MTS_IMPLEMENT_CLASS(Object,)
 NAMESPACE_END(mitsuba)
--- a/src/librender/mesh.cpp
+++ b/src/librender/mesh.cpp
@@ -857,7 +857,7 @@ MTS_VARIANT void Mesh<Float, Spectrum>::traverse(TraversalCallback *callback) {
     callback->put_parameter("vertex_normals",   m_vertex_normals);
     callback->put_parameter("vertex_texcoords", m_vertex_texcoords);
 
-    for(auto &[name, attribute]: m_mesh_attributes)
+    for (auto &[name, attribute]: m_mesh_attributes)
         callback->put_parameter(name, attribute.buf);
 }
 
@@ -887,9 +887,16 @@ MTS_VARIANT void Mesh<Float, Spectrum>::parameters_changed(const std::vector<std
     }
 }
 
+MTS_VARIANT void Mesh<Float, Spectrum>::set_grad_suspended(bool state) {
+    m_area_pmf.set_grad_suspended(state);
+}
+
 MTS_VARIANT bool Mesh<Float, Spectrum>::parameters_grad_enabled() const {
     if constexpr (ek::is_diff_array_v<Float>) {
-        return ek::grad_enabled(m_vertex_positions) ||
+        bool result = false;
+        for (auto &[name, attribute]: m_mesh_attributes)
+            result |= ek::grad_enabled(attribute.buf);
+        return result || ek::grad_enabled(m_vertex_positions) ||
                ek::grad_enabled(m_vertex_normals) ||
                ek::grad_enabled(m_vertex_texcoords);
     }
--- a/src/python/python/autodiff.py
+++ b/src/python/python/autodiff.py
@@ -179,7 +179,7 @@ def render(scene,
         if not type(spp) is tuple:
             spp = (spp, spp)
 
-        with optimizer.disable_gradients():
+        with optimizer.suspend_gradients():
             image = _render_helper(scene, spp=spp[0],
                                    sensor_index=sensor_index)
         image_diff = _render_helper(scene, spp=spp[1],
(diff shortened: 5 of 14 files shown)


include/mitsuba/core/object.h
```diff
@@ -115,6 +115,13 @@ class MTS_EXPORT_CORE Object {
      */
     virtual std::string to_string() const;
 
+    /**
+     * \brief Suspend or resume tracking of derivatives of the object and its children.
+     *
+     * This allows to stop keeping track of gradients while not recomputing intermediate values (e.g. normals, height fields...)
```
wjakob, line 121: Discussion of `normals, height fields...` is perhaps a bit specific for a documentation at the `Object` level. I'd provide a bit of context -- something like: 

> One important use case of Mitsuba 2 entails the propagation of derivatives through the full rendering process, typically via automatic differentiation. To this end, certain variants of the renderer perform all computation via Enoki's `DiffArray<T>` class, which internally builds a computation graph of all involved arithmetic. In certain situations, this is undesirable, and we need to temporarily suspend the propagation of gradients. That is the role of this function: it recursively calls `set_grad_suspended` on child objects, and it invokes `ek::set_grad_enabled` on relevant attributes.

include/mitsuba/core/distr_1d.h
```diff
@@ -233,6 +233,13 @@ template <typename Float> struct DiscreteDistribution {
         };
     }
 
+    void set_grad_suspended(bool state) {
+        if constexpr (ek::is_diff_array_v<Float>) {
```
wjakob, line 237: Here, and everywhere else in the patch, remove the `ek::is_diff_array<Float>` guard. ``ek::set_grad_suspended` expands to ..nothing.. (i.e. optimized away) when the input array is not differentiable. If you encounter any issues with that, it would be considered an Enoki bug.

src/librender/mesh.cpp
```diff
@@ -890,8 +890,19 @@ MTS_VARIANT void Mesh<Float, Spectrum>::parameters_changed(const std::vector<std
     }
 }
 
+MTS_VARIANT void Mesh<Float, Spectrum>::set_grad_suspended(bool state) {
+    if constexpr (ek::is_diff_array_v<Float>) {
```
wjakob, line 894: Here, and in other places: call parent class implementation of `set_grad_suspended` (e.g. `Shape` in this case). I suppose `Shape` should then disable the gradients of the BSDF, no?

src/librender/mesh.cpp
```diff
@@ -890,8 +890,19 @@ MTS_VARIANT void Mesh<Float, Spectrum>::parameters_changed(const std::vector<std
...
+        ek::set_grad_suspended(m_vertex_normals, state);
+        ek::set_grad_suspended(m_vertex_texcoords, state);
+        for(auto &[name, attribute]: m_mesh_attributes)
+            ek::set_grad_suspended(attribute.buf, state);
+    }
+}
+
 MTS_VARIANT bool Mesh<Float, Spectrum>::parameters_grad_enabled() const {
     if constexpr (ek::is_diff_array_v<Float>) {
+        //TODO: add mesh attributes.
```
wjakob, line 905: Not a big fan of adding TODOs (these just pile up eternally) -- let's just fix the issue when first encountered :-).

src/bsdfs/roughplastic.cpp
```diff
@@ -417,6 +417,12 @@ class RoughPlastic final : public BSDF<Float, Spectrum> {
         return oss.str();
     }
 
+    void set_grad_suspended(bool state) override {
+        if constexpr (ek::is_diff_array_v<Float>) {
+            ek::set_grad_suspended(m_external_transmittance, state);
```
wjakob, line 422: This won't ever be differntiable.

src/libcore/object.cpp
```diff
@@ -38,5 +38,7 @@ std::ostream& operator<<(std::ostream &os, const Object *object) {
     return os;
 }
 
+void Object::set_grad_suspended(bool state) { }
```
wjakob, line 41: put `state` parameter name in comment: `/* state */`, otherwise you'll get a unused parameter warning on some platforms.

src/librender/mesh.cpp
```diff
@@ -890,8 +890,19 @@ MTS_VARIANT void Mesh<Float, Spectrum>::parameters_changed(const std::vector<std
     }
 }
 
+MTS_VARIANT void Mesh<Float, Spectrum>::set_grad_suspended(bool state) {
+    if constexpr (ek::is_diff_array_v<Float>) {
+        ek::set_grad_suspended(m_vertex_positions, state);
+        ek::set_grad_suspended(m_vertex_normals, state);
+        ek::set_grad_suspended(m_vertex_texcoords, state);
+        for(auto &[name, attribute]: m_mesh_attributes)
```
wjakob, line 898: space between `for` and `(`

src/python/python/autodiff.py
```diff
@@ -179,7 +179,7 @@ def render(scene,
         if not type(spp) is tuple:
             spp = (spp, spp)
 
-        with optimizer.disable_gradients():
+        with optimizer.suspend_gradients():
```
wjakob, line 182: Nice!

wjakob (changes requested):
Hi Baptiste,

great, this looks like an excellent start! I left a few comments.

One higher-level issue that would be good to further think about (cc @Speierers) is the following: right now, you use the Optimizer class to invoke `set_grad_suspended` on all plugins that have registered parameters. Another conceptual way that it could be done is that all scene objects invoke this function recursively on children, in which case you would call `set_grad_suspended` on the scene. This could be used to catch situations where differentiable things are stored/computed in one place, and  further processed elsewhere (e.g. in the scene?).

In such a case, it would mean adding a lot more code everywhere to enable the recursive traversal -- this could potentially be avoided using a default implementation based on the already existing `traverse()` mechanism. You would then only override `set_grad_suspended` when things (fields, children) need to be suspended that aren't already covered by `traverse()`.

Best,
Wenzel

src/librender/mesh.cpp
bathal1: Right. I just wanted to do it in a separate commit :)

src/librender/mesh.cpp
bathal1: In the current implementation, `set_grad_suspended` is called on every plugin that sits in `params.hierarchy`, i.e. everything that the `traverse` mechanism went through, so in this case `set_grad_suspended` is called separately on the BSDF.

bathal1:
Hi Wenzel,
Thanks for your feedback! We discussed with @Speierers the other option you mention, of calling `scene.set_grad_suspended` and propagate through the children like the `traverse` method does. One nice thing about it is that it would allow to gracefully pause parts of the gradients, e.g. calling `set_grad_suspended`on a Shape would stop keeping track of everything related to this particular shape, while keeping track of other things in the scene graph (I'm not sure there is a use case for this though). On the other hand, this would pretty much mean reproducing the `traverse` mechanism which may not be very elegant.

src/bsdfs/roughplastic.cpp
bathal1: I'm confused about what we should stop keeping track of and what we shouldn't. All differentiable parameters end up in `params` when using `traverse(scene)` don't they? I thought the point of this mechanism was to also stop recording the "hidden" parameters that are computed from differentiable parameters and that would then be recorded.

src/librender/mesh.cpp
wjakob: Fair enough :)

src/bsdfs/roughplastic.cpp
wjakob: Generally true, but in this specific case: external_transmittance is a precomputed array that's generated using a non-differentiable algorithm.

wjakob:
Right -- the Idea I had suggested is to implement the default variant of `set_grad_suspended` (in the `Object` class) on top of the already existing `traverse`, which means that for most plugins you wouldn't even need to add any code.

Speierers:
@wjakob if I understand correctly, this would be done by writing another one of those `TraversalCallback` class where the `put_parameter` callback would suspend the gradients, right?

bathal1:
We could call the already implemented as they are `set_grad_suspended` from a custom `TraversalCallback`, which would look something like this:

```
class SuspendCallback(TraversalCallback):
    def put_parameter(...):
        # Do nothing
    def put_object(self, name, node):
        node.set_grad_suspended()
        cb = SuspendCallback(...)
        node.traverse(cb)
```

wjakob:
> @wjakob if I understand correctly, this would be done by writing another one of those TraversalCallback class where the put_parameter callback would suspend the gradients, right?
> I doubt this would work as plugins my have "intermediate" member variable that are not exposed as parameters. Which is exaclty the point of implementing a Object::set_grad_suspended().

Exactly, we'd have a specific TraversalCallback, which would get 99% of the job done. For those few cases where there are extra member variables not exposed as a parameter, you would have to then additionally override ``set_grad_suspended`` in the plugin, call the parent class version, and then manually suspend those extra fields. I don't think we will have many of them.

wjakob:
(And that TraversalCallback would be implemented in C++, not Python)

include/mitsuba/core/object.h
```diff
@@ -115,6 +115,25 @@ class MTS_EXPORT_CORE Object {
...
+     * To this end, certain variants of the renderer perform all computation via Enoki's
+     * DiffArray<T> class, which internally builds a computation graph of all involved arithmetic.
+     * In certain situations, this is undesirable, and we need to temporarily suspend the propagation of gradients.
+     * That is typically done using the Object::traverse mechanism. However, some parameters may not be exposed through
+     * in traverse, and we need to suspend them too, otherwise parts of the computation will still be recorded. This method
+     * handles such cases: its default implementation doed nothing, but classes using such "hidden" parameters may override
+     * it to properly suspend the tracking of their gradients.
+     *
+     * \param state
+     *      Whether to stop or resume recording derivatives.
```
Speierers, line 133:
```suggestion
     *      Whether to suspend or resume recording derivatives.
```

include/mitsuba/core/object.h
```diff
@@ -115,6 +115,25 @@ class MTS_EXPORT_CORE Object {
      */
     virtual std::string to_string() const;
 
+
+    /**
+     * \brief Suspend or resume tracking of derivatives.
+     *
+     * One important use case of Mitsuba 2 entails the propagation of derivatives
```
wjakob, line 122: Weird paragraph shape, each line is longer than the previous one :-). Can you wrap to 80cols?

include/mitsuba/core/object.h
```diff
@@ -115,6 +115,25 @@ class MTS_EXPORT_CORE Object {
...
+     * through the full rendering process, typically via automatic differentiation.
+     * To this end, certain variants of the renderer perform all computation via Enoki's
+     * DiffArray<T> class, which internally builds a computation graph of all involved arithmetic.
+     * In certain situations, this is undesirable, and we need to temporarily suspend the propagation of gradients.
+     * That is typically done using the Object::traverse mechanism. However, some parameters may not be exposed through
+     * in traverse, and we need to suspend them too, otherwise parts of the computation will still be recorded. This method
```
wjakob, line 128: through in -> through

src/python/python/util.py
```diff
@@ -184,3 +199,32 @@ def put_object(self, name, node):
...
+    """
+    Traverse a node of Mitsuba's scene graph and suspend/resume keeping track of derivatives.
+    """
+    from mitsuba.core import TraversalCallback
+
+    class SuspendCallback(TraversalCallback):
+        def __init__(self, node, state):
+            TraversalCallback.__init__(self)
+            self.state = state
+            self.node=node
```
wjakob, line 213: spaces around `=`.

src/textures/bitmap.cpp
```diff
@@ -592,6 +592,9 @@ class BitmapTextureImpl final : public Texture<Float, Spectrum> {
         return oss.str();
     }
 
+    void set_grad_suspended(bool state) override {
+        m_distr2d->set_grad_suspended(state);
+    }
```
wjakob, line 597: newline after `}`.

wjakob (commented):
Looks great, just a few super-minor formatting requests. Can you merge after these are applied?

PR #382, opened 2021-02-15
Normalmap traverse

belcour:
Improving the interface with python for `normalmap` and `emitter`.

First commit (cc4b6db) adds recursive `traverse` method for the `normalmap` plugin that enables to access the BSDF below the `normalmap` in the hierarchy. This enables to select the albedo of a diffuse BRDF for optimization purposes.

Second commit (ff221e4) exposes a method to change the `AnimatedTransform` of an `emitter`.

### DIFF
--- a/src/bsdfs/normalmap.cpp
+++ b/src/bsdfs/normalmap.cpp
@@ -156,6 +156,11 @@ public:
         return result;
     }
 
+    void traverse(TraversalCallback *callback) override {
+       callback->put_object("normalmap", m_normalmap.get());
+       callback->put_object("nested_bsdf", m_nested_bsdf.get());
+    }
+
     std::string to_string() const override {
         std::ostringstream oss;
         oss << "NormalMap[" << std::endl


src/bsdfs/normalmap.cpp
```diff
@@ -155,6 +155,11 @@ class NormalMap final : public BSDF<Float, Spectrum> {
         result.t = cross(result.n, result.s);
         return result;
     }
+    
+    void traverse(TraversalCallback *callback) override {
+        callback->put_object("normalmap", m_normalmap.get());
+        m_nested_bsdf->traverse(callback);
```
Speierers, line 161: Here I would rather use the `put_object` method as for other plugins:
```
        callback->put_object("nested_bsdf", m_nested_bsdf.get());
```

Speierers:
Hi @belcour ,

Thanks for opening this PR. 

Regarding your first commit, this change was already made in the upcoming refactored codebase so I don't think it is necessary to push it onto `master`.

Support for `AnimatedTransform` and accessing/setting transforms of scene objects is on our roadmap and will require significant changes to the codebase. Therefore I would rather wait for a more general solution that also fixes it for other objects (e.g. `Shape`).

belcour:
No problem Sebastien.

For the last part, do you plan to move the AnimatedTransform inside the base class `Object` ?

Speierers:
> For the last part, do you plan to move the AnimatedTransform inside the base class Object ?

Most likely not, as `Object` doesn't necessarily represent a "geometric" scene object (e.g. `Film`, `Texture`, `BSDF`, ...)

belcour:
Ok. Then I would suggest not to overlook the modification I am suggesting then: since `Endpoints` are separate objects from other scene elements such as `Shape` it makes sense to treat them separately. Also it is kind of an easy fix (compared to adding another layer of object heritancy).

Speierers:
Also we will likely use the `traverse` mechanism for transforms as well as those could be differentiable.

wpalfi:
> Regarding your first commit, this change was already made in the upcoming refactored codebase so I don't think it is necessary to push it onto `master`.

You are not alone @belcour. I just stumbled upon that.

@Speierers the fix is trivial, so why not merge it into the master? Or do you plan to release the refactored version soon?

Speierers:
Hi @wpalfi,
Happy to push those 4 lines in `normalmap.cpp` but I would rather not merge the other changes included in this PR. @belcour would you mind isolating this commit?

Our hope is to release the refactored version soon, but I can't give you a more specific ETA for now unfortunately.

belcour:
Hi @Speierers,

I reset the branch to a single commit and used `put_object` instead of the traverse call. Hope this is fine for you.

Speierers:
Thanks @belcour !

PR #384, opened 2021-02-20
Emissive Media Rendering in Volpathmis

### LINKED ISSUE #312: [❔ other question] Implementing Emissive Medium Rendering
## Summary

I would like to implement rendering of an emissive medium (that is also possibly absorbing/scattering), but I'm not sure what the best place to start is. From what I can tell, the `volpath` integrator assumes that mediums are attenuating/scattering and the base class `medium` is implemented with this assumption.

My understanding is that, at a minimum, the `medium` class would need to be updated to account for emission and a new volumetric integrator would need to be included that can correctly sample emissive volumes. I don't know if you have anyone working on this feature, but I would be really grateful if you could point me towards any resources or aspects of the code that might help with implementing this feature.

Microno95:
This PR adds emissive event sampling in the null-scattering `volpathmis` integrator for sampling the VRE.

## Description

Adds emissive volume support by making the following changes:

- `src/libcore/xml.h`: Added an additional boolean check for attributes named `radiance`, this is treated as equivalent to nodes being within an emitter in `create_texture_from_spectrum` and `create_texture_from_rgb`.
- `include/mitsuba/render/interaction.h`: Added attribute `radiance` to `MediumInteraction` so that the radiance at the sampled point is stored in the interaction following the same convention as other attributes (eg. sigma_t).
- `include/mitsuba/render/medium.h`: Added virtual method `get_radiance(mi, active)` for sampling radiance at a given interaction and included this in the enoki_vcall_method declarations.
- `src/librender/medium.cpp`: Added line `mi.radiance = get_radiance(mi, valid_mi)` for sampling and storing the radiance in the interaction.
- `src/media/homogeneous.cpp` and `src/media/heterogeneous.cpp`: Added `m_radiance` as an attribute and added initialisation of the attribute in the constructor via the `radiance` property. Default value set to no radiance.
- `src/integrators/volpathmis.cpp`: Added sampling of the `emission` event in `sample` when sampling the RTE. Updated probabilities for each event and added the null-scatter event probability calculation to `sample_emitter` to ensure consistency.
- `src/librender/python/medium_v.cpp`: Added Python interface for `get_radiance`.
- `src/librender/python/interaction_v.cpp`: Added attribute `radiance` to Python attribute declarations.
- `include/mitsuba/python/docstr.h`: Added docstrings for python declarations of added attributes and methods.

Fixes #312

## Testing

- The updated codebase compiles with no new warnings.
- Rendered the entirety of the `participating_media` tests without any error. Although I haven't run the automated python tests, the results appear to be correct albeit with increased variance due to the termination of rays when emission events are sampled - this can be improved by turning off emission event sampling when radiance is zero, but this adds code complexity so I've omitted it for this PR.
- Generated several emission test variants in `participating_media` which the integrator renders. These will probably need a separate PR for the mitsuba resources repo - additionally, these would need reference renders to be generated.
- Tested compilation for all `scalar_` and `llvm_` single precision float modes.

## Checklist:

- [ ] My code also compiles for `cuda_*` and `llvm_*` variants.
- [x] I have made corresponding changes to the documentation
- [x] I cleaned the commit history and removed any "Merge" commits

## Known Issues and Notes

- The implementation does not use the analogue probabilities as described in most null-scattering papers for 2 reasons:
	1. The radiance herein is assumed to not be the product of `sigma_a` and `L_e`, but rather simply `L_e` therefore the probability of emission should be proportional to the radiance.
	2. The scattering probability is left proportional to `sigma_t` for the reason that when proportional to `sigma_s` I was observing a reduced throughput in spectral rendering modes when the term `index_spectrum(sigma_s, channel)` was zero while non-zero for other channels. This could still be an issue with the current setup so might be worth looking into.
- Radiance in `spectral` modes is ~50 times brighter than `mono` and `rgb` modes when specified as `rgb` and ~2-4 times dimmer when specified as a `spectrum`. I am investigating this by looking at how the spectral conversion operates, might also be an event probability sampling issue similar to the scattering probability above.
- Could not test compilation for `cuda_` variants as I am using Windows Subsystem for Linux and the code refuses to compile under Windows with `cuda_` modes. Shall I create a bug report for this?
- Could not test compilation for `double` modes as the compiler threw an error regarding pointer type mismatches.
- Could not test for correct rendering under the `llvm_` modes as the code would simply not render (left it running for 8 hours overnight but the llvm computation graph still wasn't constructed)


wjakob:
Hi @Microno95,

it says above that this PR was closed by me. Just to clarify, this is not an issue with your work but rather something that happened automatically as a consequence of temporarily taking the "next" branch offline. The reasoning is explained here: https://github.com/mitsuba-renderer/mitsuba2/blob/next/README.md

Can you let me know about your plans with this PR? Is it ready to be reviewed? We can give you private access to the "next" branch if you would like to continue working on it. Obviously the plans for this branch is that it will be merged into 'master' at some point. We just don't know when this will be as this next version of the Mitsuba renderer is now effectively in the category "unpublished research".

Best,
Wenzel

Microno95:
Hi @wjakob,

Ah that makes sense, I'm glad to hear it's not about my work 😃 

The basic implementation works; currently my priority is to implement NEE sampling of emissive volumes and to quash a weird discrepancy related to energy differences between `rgb` and `spectral` modes. If I had access to the `next` branch I could continue working on those and also ensure that as the `next` branch is updated, this PR does not introduce any bugs.

Best,
Ekin

wjakob:
OK, I will add you. Please do not share it with others for the time being.

wjakob:
(You should then be able to open a new pull request there)

PR #386, opened 2021-02-24
Blender refactoring

bathal1:
## Description

This PR introduces some bug fixes to the blender exporter, as well as a reimplementation of the `Bitmap::split method`.

Calling `Bitmap::split` now splits into bitmaps according to the prefix, and not only the pixel format. To that end, I added the possibility to provide a list of channel names to the `Bitmap` constructor.

A unit test was added to check the new behavior.

### DIFF
--- a/include/mitsuba/core/bitmap.h
+++ b/include/mitsuba/core/bitmap.h
@@ -182,17 +182,17 @@ public:
     // ======================================================================
 
     /**
-     * \brief Create a bitmap of the specified type and allocate
-     * the necessary amount of memory
+     * \brief Create a bitmap of the specified type and allocate the necessary
+     * amount of memory
      *
      * \param pixel_format
      *    Specifies the pixel format (e.g. RGBA or Luminance-only)
      *
      * \param component_format
-     *    Specifies how the per-pixel components are encoded
-     *    (e.g. unsigned 8 bit integers or 32-bit floating point values).
-     *    The component format struct_type_v<Float> will be translated to the
-     *    corresponding compile-time precision type (Float32 or Float64).
+     *    Specifies how the per-pixel components are encoded (e.g. unsigned 8
+     *    bit integers or 32-bit floating point values). The component format
+     *    struct_type_v<Float> will be translated to the corresponding
+     *    compile-time precision type (Float32 or Float64).
      *
      * \param size
      *    Specifies the horizontal and vertical bitmap size in pixels
@@ -200,6 +200,10 @@ public:
      * \param channel_count
      *    Channel count of the image. This parameter is only required when
      *    \c pixel_format = \ref PixelFormat::MultiChannel
+     *
+     * \param channel_names
+     *    Channel names of the image. This parameter is optional, and only used
+     *    when \c pixel_format = \ref PixelFormat::MultiChannel
      *
      * \param data
      *    External pointer to the image data. If set to \c nullptr, the
@@ -209,6 +213,7 @@ public:
            Struct::Type component_format,
            const Vector2u &size,
            size_t channel_count = 0,
+           const std::vector<std::string> &channel_names = {},
            uint8_t *data = nullptr);
 
     /**
@@ -585,7 +590,7 @@ public:
      virtual ~Bitmap();
 
      /// Rebuild the 'm_struct' field based on the pixel format etc.
-     void rebuild_struct(size_t channel_count = 0);
+     void rebuild_struct(size_t channel_count = 0, const std::vector<std::string> &channel_names = {});
 
      /// Read a file from a stream
      void read(Stream *stream, FileFormat format);
--- a/src/bsdfs/measured_polarized.cpp
+++ b/src/bsdfs/measured_polarized.cpp
@@ -271,7 +271,7 @@ public:
 
                 /* Invalid configurations such as transmission directions are encoded as NaNs.
                    Make sure these values don't end up in the interpolated value. */
-                ek::masked(value, ek::any(isnan(value(0,0)))) = 0.f;
+                ek::masked(value, ek::any(ek::isnan(value(0,0)))) = 0.f;
 
                 // Make sure intensity is non-negative
                 value(0,0) = ek::max(0.f, value(0,0));
--- a/src/films/hdrfilm.cpp
+++ b/src/films/hdrfilm.cpp
@@ -193,9 +193,9 @@ public:
         channels_sorted.push_back("G");
         channels_sorted.push_back("B");
         std::sort(channels_sorted.begin(), channels_sorted.end());
-        for (size_t i = 1; i < channels.size(); ++i) {
-            if (channels[i] == channels[i - 1])
-                Throw("Film::prepare(): duplicate channel name \"%s\"", channels[i]);
+        for (size_t i = 1; i < channels_sorted.size(); ++i) {
+            if (channels_sorted[i] == channels_sorted[i - 1])
+                Throw("Film::prepare(): duplicate channel name \"%s\"", channels_sorted[i]);
         }
 
         m_storage = new ImageBlock(m_crop_size, channels.size());
--- a/src/films/hdrfilm.cpp
+++ b/src/films/hdrfilm.cpp
@@ -256,7 +256,7 @@ public:
         ref<Bitmap> source = new Bitmap(m_channels.size() != 5 ? Bitmap::PixelFormat::MultiChannel
                                                                : Bitmap::PixelFormat::XYZAW,
                           struct_type_v<ScalarFloat>, m_storage->size(), m_storage->channel_count(),
-                          (uint8_t *) storage.data());
+                          m_channels, (uint8_t *) storage.data());
 
         if (raw)
             return source;
@@ -310,8 +310,6 @@ public:
                         dest_field.name = m_channels[i];
                         break;
                 }
-
-                source_field.name = m_channels[i];
             }
         }
 
--- a/src/integrators/aov.cpp
+++ b/src/integrators/aov.cpp
@@ -91,7 +91,7 @@ public:
 
             if (item[1] == "depth") {
                 m_aov_types.push_back(Type::Depth);
-                m_aov_names.push_back(item[0]);
+                m_aov_names.push_back(item[0] + ".T");
             } else if (item[1] == "position") {
                 m_aov_types.push_back(Type::Position);
                 m_aov_names.push_back(item[0] + ".X");
--- a/src/integrators/aov.cpp
+++ b/src/integrators/aov.cpp
@@ -170,7 +170,7 @@ public:
         for (size_t i = 0; i < m_aov_types.size(); ++i) {
             switch (m_aov_types[i]) {
                 case Type::Depth:
-                    *aovs++ = si.t;
+                    *aovs++ = ek::select(si.is_valid(), si.t, 0.f);
                     break;
 
                 case Type::Position:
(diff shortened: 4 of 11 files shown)


include/mitsuba/core/bitmap.h
```diff
@@ -182,33 +182,34 @@ class MTS_EXPORT_CORE Bitmap : public Object {
...
+     * \param channel_count Channel count of the image. This parameter is only
+     *    required when \c pixel_format = \ref PixelFormat::MultiChannel
      *
-     * \param data
-     *    External pointer to the image data. If set to \c nullptr, the
+     * \param channel_names Channel names of the image. This parameter is
+     *    optional, and only used when \c pixel_format = \ref
+     *    PixelFormat::MultiChannel
+     *
+     * \param data External pointer to the image data. If set to \c nullptr, the
```
Speierers, line 205: For consistency, I would prefer keeping the new line after the parameter name (new line for the description)

src/bsdfs/measured_polarized.cpp
```diff
@@ -271,7 +271,7 @@ class MeasuredPolarized final : public BSDF<Float, Spectrum> {
 
                 /* Invalid configurations such as transmission directions are encoded as NaNs.
                    Make sure these values don't end up in the interpolated value. */
-                ek::masked(value, any(isnan(value(0,0)))) = 0.f;
+                ek::masked(value, ek::any(ek::isnan(value(0,0)))) = 0.f;
```
Speierers, line 274: Good catch 😄

src/integrators/aov.cpp
```diff
@@ -91,7 +91,7 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
 
             if (item[1] == "depth") {
                 m_aov_types.push_back(Type::Depth);
-                m_aov_names.push_back(item[0]);
+                m_aov_names.push_back(item[0] + ".T");
```
Speierers, line 94: That channel gets put into `root` as it doesn't have a `.` is that correct? Shouldn't we fix this behaviour in `Bitmap::split()` directly instead?

src/libcore/bitmap.cpp
```diff
@@ -64,7 +64,8 @@ extern "C" {
 NAMESPACE_BEGIN(mitsuba)
 
 Bitmap::Bitmap(PixelFormat pixel_format, Struct::Type component_format,
-               const Vector2u &size, size_t channel_count, uint8_t *data)
+               const Vector2u &size, size_t channel_count,
+               std::vector<std::string> channel_names, uint8_t *data)
```
Speierers, line 68:
```suggestion
               const std::vector<std::string> &channel_names, uint8_t *data)
```

include/mitsuba/core/bitmap.h
```diff
@@ -182,33 +182,34 @@ class MTS_EXPORT_CORE Bitmap : public Object {
...
      *    implementation will allocate memory itself.
      */
     Bitmap(PixelFormat pixel_format,
            Struct::Type component_format,
            const Vector2u &size,
            size_t channel_count = 0,
+           std::vector<std::string> channel_names = {},
```
Speierers, line 212:
```suggestion
           const std::vector<std::string> &channel_names = {},
```

src/films/hdrfilm.cpp
```diff
@@ -256,7 +256,7 @@ class HDRFilm final : public Film<Float, Spectrum> {
         ref<Bitmap> source = new Bitmap(m_channels.size() != 5 ? Bitmap::PixelFormat::MultiChannel
                                                                : Bitmap::PixelFormat::XYZAW,
                           struct_type_v<ScalarFloat>, m_storage->size(), m_storage->channel_count(),
-                          (uint8_t *) storage.data());
+                          m_channels, (uint8_t *) storage.data());
```
Speierers, line 259: Should we fix the `prepare` method above as well?

src/libcore/bitmap.cpp
```diff
@@ -587,107 +599,87 @@ std::vector<std::pair<std::string, ref<Bitmap>>> Bitmap::split() const {
...
-                            : PixelFormat::XYZ)
-                   : (has_a ? PixelFormat::YA
-                            : PixelFormat::Y)),
+                    ? (has_a ? PixelFormat::RGBA
+                            : PixelFormat::RGB)
+                    : (has_xyz
+                    ? (has_a ? PixelFormat::XYZA
+                                : PixelFormat::XYZ)
+                    : (has_a ? PixelFormat::YA
+                                : PixelFormat::Y)),
```
Speierers, line 641: Fix indentation

src/shapes/blender.cpp
```diff
@@ -306,6 +323,7 @@ class BlenderMesh final : public Mesh<Float, Spectrum> {
         }
 
         set_children();
+
```
Speierers, line 326: Remoave this empty line

src/shapes/blender.cpp
```diff
@@ -306,6 +323,7 @@ class BlenderMesh final : public Mesh<Float, Spectrum> {
...
     }
 
     MTS_DECLARE_CLASS()
```
Speierers, line 329: Add an empty line at the end of this file

Speierers (changes requested):
Thanks for all those changes @bathal1. Looks pretty good to me. Here are a few feedbacks.

src/integrators/aov.cpp
bathal1: I'm not sure it's a good idea to let `Bitmap::split` handle these cases. I think it would be better to make sure we don't create `Bitmap`s with an ambiguous layout altogether, and assume that in `split`.

src/films/hdrfilm.cpp
bathal1: What would need fixing there?

bathal1: Ah, you mean the sorting that's broken? Right, I had forgotten about that. I will commit it.

Speierers: THanks!

src/integrators/aov.cpp
Speierers: Could `split` output a warning in this case?

src/films/hdrfilm.cpp
```diff
@@ -310,8 +310,6 @@ class HDRFilm final : public Film<Float, Spectrum> {
                         dest_field.name = m_channels[i];
                         break;
                 }
-
-                source_field.name = m_channels[i];
```
wjakob, line 314: Curious about this deletion.

src/integrators/aov.cpp
```diff
@@ -170,7 +170,7 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
         for (size_t i = 0; i < m_aov_types.size(); ++i) {
             switch (m_aov_types[i]) {
                 case Type::Depth:
-                    *aovs++ = si.t;
+                    *aovs++ = ek::select(si.is_valid(), si.t, 0.f);
```
wjakob, line 173: Great catch!

src/films/hdrfilm.cpp
bathal1: The list of channel names can be passed to the `Bitmap` constructor (see L259).

src/libcore/bitmap.cpp
```diff
@@ -185,11 +197,11 @@ void Bitmap::rebuild_struct(size_t channel_count) {
     m_struct = new Struct();
     for (auto ch: channels) {
         uint32_t flags = +Struct::Flags::None;
-        if (ch != "A" && ch != "W" && m_srgb_gamma)
+        if ((ch != "A" || m_pixel_format == PixelFormat::MultiChannel) && ch != "W" && m_srgb_gamma)
```
wjakob, line 200: I don't understand this logic here, and it seems very repeated across lines. can you clarify/refactor?

wjakob:
General comment: there are extensive changes to `rebuild_struct` that I would like to understand better. But the commit message does not provide enough context for this.

src/films/hdrfilm.cpp
wjakob: ok!

src/libcore/bitmap.cpp
bathal1: With the new logic for creating multichannel Bitmaps by specifying the field names, if the Bitmap contains a field "A", it will be flagged as the alpha channel. Imagine now you have a multichannel bitmap with fields `[R, G, B, A, im.R, im.G, im.B, im.A]`.

When calling `split` on this Bitmap, a `StructConverter` will be created, taking as argument the "main" Bitmap, and the split one we're currently creating. For the `im` Bitmap, the field `im.A` will be flagged as alpha (as it should). However the `StructConverter` will raise an error because the alpha channels in the two structs are mismatched (`A` vs `im.A`).

This was not an issue before because we usually call `film.bitmap().split()` and the `HDRFilm::Bitmap` creates a multichannel bitmap (in case it has aovs) with generic channel names and renames them afterwards, hence not flagging any channel as the alpha in the multichannel bitmap. 

This modification to `rebuild_struct` aims at working around that, by simply not flagging any channel as alpha in the multichannel case, since the notion of alpha in this case is ambiguous and we don't do anything alpha related with it anyway. 

The same issue arises for the `W` channel. For now, the workaround for it is to fail if a channel called W is encountered in the `split` method, as the only time this channel should be added is manually in the `Integrator::render` method.

src/integrators/aov.cpp
bathal1: I'm not sure that's desirable, what would happen then if we created a Multichannel bitmap with only the root node? e.g. `[T,U,V]` as channels. In this case there's no way to tell whether these channels should really be there or not.

src/libcore/bitmap.cpp
wjakob: Ok, that sounds reasonable to me. Can you refactor the code to move some conditionals at the top of the loop? (e.g. `bool is_alpha = ch == "A" && m_pixel_format != PixelFormat::MultiChannel;`. Please also add this explanation to the commit message.

bathal1: Sure. Do you also agree with the current handling of the "W" case?

wjakob: it's been too long since I wrote that code, so I probably don't have the same level of insight that you hold now. As far as I can remember, we can have multichannel images with a weight channel, and it's important to keep track of that and at some point divide by the weight as the image is exposed. This is the standard situation when rendering AOVs I suppose. As long as that is working, I don't see an issue.

Speierers: Regarding the `W` channel, I think it is fair to require the weight channel to already be applied before calling `split`, which is what we have at the moment. Therefore I am in favor of Baptiste's changes where the `split` method fails when encountering a `W` channel.

wjakob:
Makes sense!

On Thu, Feb 25, 2021, at 12:57 PM, Sebastien Speierer wrote:
> 

> *@Speierers* commented on this pull request.

> 
> In src/libcore/bitmap.cpp <https://github.com/mitsuba-renderer/mitsuba2/pull/386#discussion_r582774715>:

> > @@ -185,11 +197,11 @@ void Bitmap::rebuild_struct(size_t channel_count) {
     m_struct = new Struct();
     for (auto ch: channels) {
         uint32_t flags = +Struct::Flags::None;
-        if (ch != "A" && ch != "W" && m_srgb_gamma)
+        if ((ch != "A" || m_pixel_format == PixelFormat::MultiChannel) && ch != "W" && m_srgb_gamma)

src/integrators/aov.cpp
Speierers: So if I understand correctly, the current behaviour will be that all channels without a prefix will be placed in `root`. Is that correct?

bathal1: yes.

merlinND:
This issue is relevant to the `Bitmap::split` method: #389

PR #391, opened 2021-03-05
Adding textured roughness to the roughplastic material

belcour:
Adding textured roughness to the `roughplastic` plugin.
This is done similarly to the `roughconductor` plugin.

### DIFF
--- a/src/bsdfs/roughplastic.cpp
+++ b/src/bsdfs/roughplastic.cpp
@@ -165,15 +165,20 @@ public:
 
         m_nonlinear = props.bool_("nonlinear", false);
 
-        mitsuba::MicrofacetDistribution<ScalarFloat, Spectrum> distr(props);
-        m_type = distr.type();
-        m_sample_visible = distr.sample_visible();
-
-        if (distr.is_anisotropic())
-            Throw("The 'roughplastic' plugin currently does not support "
-                  "anisotropic microfacet distributions!");
-
-        m_alpha = distr.alpha();
+        /// Load Microfacet distribution
+        m_type = MicrofacetType::Beckmann;
+        if (props.has_property("distribution")) {
+            std::string distr = string::to_lower(props.string("distribution"));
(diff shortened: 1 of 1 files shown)


Speierers:
Hi @belcour ,

Thanks for opening this PR and looking into this.

IIRC the reason for not supporting textured roughness in this plugin is that the precompute tables `m_external_transmittance` and `m_internal_transmittance` depend on the `alpha` and would therefore need to be spacially varying as well.

We currently do not have a good way to deal with this problem, and are open to suggestions!

belcour:
Hi @Speierers,

I see your point. Right now, my way of handling it is to feed the mean value of the roughness texture to the preintegrated transmittance and reflectance. Visually, the dieletric coating's roughness is the most important cue.

Mitsuba 0.6 did precompute the tabulated versions using 3D data tables. Anything preventing to use the 'old' method?

A more proper way would be to stochastically evaluate the rough refractions and diffuse bounce in the BSDF. Given that the outcoming importance will be angularly very low frequency, that shouldn't be too much of an issue with respect to increasing variance in the integrator.

Speierers:
Hi @belcour,

Those are indeed viable solutions. We orgininally thought of fitting polynomials to this high dimentional table that could then be evaluated analytically. This would be more elegant than precomputing a table and more accurate than the solution you proposed here.

Any interest in giving this a shot?

belcour:
Hi @Speierers,

Sure, I can take a lot at that (might take a while though). Still, the polynomial might not be the most efficient way to handle this given the dimensionality of the data.

waytobehigh:
Hi @Speierers!

Is there a way to compute `m_external_transmittance` and `m_internal_transmittance` on the fly without pre-calculation? Non-painful way is preferred :)

wjakob:
It would be nice if there was. Do let us know if you can think of something @waytobehigh ;-). A student at EPFL is also currently looking into this, we'll see..

PR #399, opened 2021-03-16
Remove TBB dependency and use the enoki-thread library instead

Speierers:
This PR attempts to get rid of the TBB dependency in Mitsuba, replacing it by the `enoki-thread` library.


include/mitsuba/core/tls.h
```diff
@@ -1,3 +1,4 @@
+#if 0
```
wjakob, line 1: remove entirely then?

src/libcore/bitmap.cpp
```diff
@@ -796,26 +792,10 @@ void Bitmap::write(Stream *stream, FileFormat format, int quality) const {
...
-            bitmap->write(path, format, quality);
-            return nullptr;
-        }
-    };
-
-    WriteTask *t = new (tbb::task::allocate_root())
-        WriteTask(this, path_, format_, quality_);
-    tbb::task::enqueue(*t);
+void Bitmap::write_async(const fs::path &path, FileFormat format, int quality) const {
+    ek::do_async([&](){
```
wjakob, line 796: That's a nice size reduction :-)

src/libcore/python/main.cpp
```diff
@@ -73,25 +70,26 @@ PYBIND11_MODULE(core_ext, m) {
...
         fr->append(base_path);
 
-    /* Holds a pointer to the current TBB task scheduler to enable
-       later changes via set_thread_count(). Cleaned up in
-       \ref cleanup_callback. */
-    std::unique_ptr<tbb::task_scheduler_init> *task_scheduler_holder =
-        new std::unique_ptr<tbb::task_scheduler_init>(
-            new tbb::task_scheduler_init());
-    m.def(
-        "set_thread_count",
```
wjakob, line 83: Would be nice to expose the `Thread::set_thread_count` function (or `pool_set_size` from enoki-thread) somehow. We may not always want to run the Python bindings with all threads

src/libcore/thread.cpp
```diff
@@ -85,12 +91,30 @@ class WorkerThread : public Thread {
 
 std::atomic<uint32_t> WorkerThread::m_counter{0};
 
+struct ThreadRegistrator {
```
wjakob, line 94: Registrator sounds weird. Perhaps "Notifier"?

src/libcore/thread.cpp
```diff
@@ -85,12 +91,30 @@ class WorkerThread : public Thread {
...
+    ThreadRegistrator() {
+        // Do not register the main thread
+        if (m_counter > 0)
+            Thread::register_external_thread("worker");
```
wjakob, line 98: Give it a number? (with `snprintf("wrk_%02i", ...)`)

src/libcore/thread.cpp
```diff
@@ -19,10 +18,17 @@
 #  include <windows.h>
 #endif
 
+// TODO
```
wjakob, line 21: What was the issue here? IIRC this is only called on linux.

src/libcore/thread.cpp
wjakob: Actually, can we not just get rid of this class completely now?

src/libcore/thread.cpp
```diff
@@ -405,14 +429,12 @@ void Thread::dispatch() {
...
-
 }
 
 void Thread::exit() {
     Log(Debug, "Thread \"%s\" has finished", d->name);
     d->running = false;
-    Assert(*self == this);
-    ThreadLocalBase::unregister_thread();
+    Assert(self == this);
     dec_ref();
```
wjakob, line 438: add `self = nullptr` here?

src/libcore/tls.cpp
```diff
@@ -1,5 +1,6 @@
+#if 0
 #include <mitsuba/core/tls.h>
```
wjakob, line 2: RIP tls.cpp

src/libcore/xml.cpp
```diff
@@ -972,66 +971,63 @@ static ref<Object> instantiate_node(XMLParseContext &ctx, const std::string &id)
...
-                    props.set_object(kv.first, obj, false);
-                } else if (children.size() == 1) {
-                    props.set_object(kv.first, children[0], false);
-                } else {
-                    int ctr = 0;
-                    for (auto c : children)
-                        props.set_object(kv.first + "_" + std::to_string(ctr++), c, false);
+        ek::parallel_for(
+            ek::blocked_range<uint32_t>(0u, (uint32_t) named_references.size(), task_size),
+            [&](const ek::blocked_range<uint32_t> &range) {
```
wjakob, line 997: Getting a bit too nested. You could declare the lambda as a variable above and then call it from `ek:parallel_for`

src/libcore/xml.cpp
```diff
@@ -972,66 +971,63 @@ static ref<Object> instantiate_node(XMLParseContext &ctx, const std::string &id)
...
+                        // sub-objects
+                        std::vector<ref<Object>> children = obj->expand();
+                        if (children.empty()) {
+                            props.set_object(kv.first, obj, false);
+                        } else if (children.size() == 1) {
+                            props.set_object(kv.first, children[0], false);
+                        } else {
+                            int ctr = 0;
+                            for (auto c : children)
+                                props.set_object(kv.first + "_" +
```
wjakob, line 1015: ugly indentation :)

src/libcore/xml.cpp
```diff
@@ -972,66 +971,63 @@ static ref<Object> instantiate_node(XMLParseContext &ctx, const std::string &id)
...
+            [&](const ek::blocked_range<uint32_t> &range) {
+                ScopedSetThreadEnvironment set_env(env);
+
+                for (uint32_t i = range.begin(); i != range.end(); ++i) {
+                    auto &kv = named_references[i];
+                    try {
+                        ref<Object> obj = instantiate_node(ctx, kv.second);
+
+                        // Give the object a chance to recursively expand into
+                        // sub-objects
```
wjakob, line 1006: ugly indentation :)

src/mitsuba/mitsuba.cpp
```diff
@@ -223,8 +219,8 @@ int main(int argc, char *argv[]) {
 
         // Initialize Intel Thread Building Blocks with the requested number of threads
         if (*arg_threads)
-            __global_thread_count = arg_threads->as_int();
-        tbb::task_scheduler_init scheduler(std::max(1, (int) __global_thread_count));
+            Thread::set_thread_count(std::max(1, arg_threads->as_int()));
+        jit_llvm_set_thread_count(Thread::thread_count());
```
wjakob, line 223: Is this perhaps unnecessary? My guess would be taht both of those methods will update the enoki-jit thread pool size, no?

wjakob (commented):
Looking great! Here are a few small comments.

include/mitsuba/core/tls.h
Speierers: Yes indeed

src/libcore/python/main.cpp
Speierers: It's already exposed by `enoki`:
https://github.com/wjakob/enoki/blob/5b5d5c9bbb45ab616054824217a1123d8f2d5e35/src/python/main.cpp#L207

Although we could rename that binding into `pool_set_size`.

src/mitsuba/mitsuba.cpp
Speierers: I will clean up that logic

src/libcore/python/main.cpp
wjakob: We could remove the `llvm_` prefix.

src/libcore/thread.cpp
Speierers: This is just the "prefix", could change this to `wrk_` if you prefer that.

src/libcore/thread.cpp
Speierers: `static thread_local ThreadRegistrator;` will register all the external threads. I still we still want that, no?

src/libcore/thread.cpp
```diff
@@ -85,12 +88,30 @@ class WorkerThread : public Thread {
...
+    }
+    ~ThreadNotifier() {
+        if (self)
+            Thread::unregister_external_thread();
+        m_counter--;
+    }
+    static std::atomic<uint32_t> m_counter;
+}
+
+static thread_local ThreadNotifier;
```
wjakob, line 106: I am confused by this line. TLS behavior is largely plaform-dependent. As such, I don't think we can expect that this will be correctly initialized everywhere in such an anonymous variable-- rather, some kind of access to this variable will be needed, which will then check if the object has been constructed. If you know more than I do, then I'd be curious to learn more.

src/libcore/thread.cpp
```diff
@@ -13,6 +11,11 @@
 // Required for native thread functions
 #if defined(__LINUX__)
 #  include <sys/prctl.h>
+#  ifdef _SC_NPROCESSORS_CONF
+#    define NPROCESSORS_COUNT sysconf(_SC_NPROCESSORS_CONF)
```
wjakob, line 15: I still don't understand why this has to be declared at the top. It always worked on linux, no?

wjakob:
A few more comments. Also, is the deadlock resolved? Curious what the issue was.

Speierers:
Haven't managed to solved the issue of the deadlock yet. We used to use `tbb::isolate` for this. Basically it could be that while instantiating some nested nodes, we instanciate a node of the same type, locking on the same mutex yielding a deadlock.

wjakob:
I've thought of a different way for doing all of this, see slack.

src/libcore/thread.cpp
Speierers: Right, we need to include `#include <unistd.h>` then.

src/libcore/thread.cpp
Speierers: Would you like this better?
```
static thread_local ThreadNotifier notifier{};
```

wjakob: No, I don't think that really addresses the issue in a cross-platform way. The problem is that `notifier` will not automatically get called if a new thread launches (it seems to me you are assuming that happens everywhere, correct?)

include/mitsuba/render/kdtree.h
```diff
@@ -542,6 +535,125 @@ template <typename BoundingBox_, typename Index_, typename CostModel_,
     /*                    Build-related data structures                     */
     /* ==================================================================== */
 
+    /* Append-only concurrent vector, whose storage is arranged into slices
```
wjakob, line 538: A small request: A lot of this code technicaly doesn't depend on any KDTree template parameters. Can you move it to a `detail::` namespace? That includes `OrderedChunkAllocator` and `concurrent_vector`, maybe others as well. For `concurrent_vector`, it should probably renamed `ConcurrentVector` for consistency.

wjakob: `Index` and `Size` should remain `uint32_t` types here, because this code really will not work anymore if you change to `uint64_t`

include/mitsuba/render/kdtree.h
```diff
@@ -1103,20 +1205,21 @@ template <typename BoundingBox_, typename Index_, typename CostModel_,
...
+            BuildTask right_task = BuildTask(
+                m_ctx, children + 1, std::move(partition.right_indices),
                 right_bounds, partition.right_bounds, m_depth + 1,
                 m_bad_refines, &right_cost);
 
-            set_ref_count(3);
-            spawn(left_task);
-            spawn(right_task);
-            wait_for_all();
+            Task *left_ek_task = ek::do_async([&](){ left_task.execute(); });
```
wjakob, line 1218: There are two small ways of improving this code:

1. We don't need to call `do_async` for both tasks. Instead, we can execute one on the current thread (which is waiting) and push another one onto the job queue.
2. We should switch over to serial execution and completely avoid the `do_async` once the tasks become small enough. (GRAIN_SIZE sounds reasonable)

PR #406, opened 2021-03-31
Add scale parameter to blackbody emitter [feature]

alhirzel:
## Description

This PR adds the `scale` parameter back to the `blackbody` emitter. This feature is ported from the old Mitsuba and is useful when scenes are not expressed in meters, or when simulating a gray body.

## Testing

*Please describe the tests that you added to verify your changes.*

## Checklist:

*Please make sure to complete this checklist before requesting a review.*

- [ ] My code also compiles for `gpu_*` and `packet_*` variants. If you can't test this, please leave below
- [x] I have made corresponding changes to the documentation

### DIFF
--- a/src/spectra/blackbody.cpp
+++ b/src/spectra/blackbody.cpp
@@ -11,8 +11,8 @@ NAMESPACE_BEGIN(mitsuba)
 sRGB D65 spectrum (:monosp:`blackbody`)
 ---------------------------------------
 
-This is a black body radiation spectrum for a specified temperature
-And therefore takes a single :monosp:`float`-valued parameter :paramtype:`temperature` (in Kelvins).
+This is a black body radiation spectrum for a specified temperature.
+The temperature of the black body in degrees Kelvin is given by the :paramtype:`temperature` parameter.
 
 This is the only spectrum type that needs to be explicitly instantiated in its full XML description:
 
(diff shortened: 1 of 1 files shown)


alhirzel:
I removed WIP but haven't run the tests formally yet. What level of test would be expected for this change?

src/spectra/blackbody.cpp
```diff
@@ -53,6 +61,10 @@ class BlackBodySpectrum final : public Texture<Float, Spectrum> {
 
     BlackBodySpectrum(const Properties &props) : Texture(props) {
         m_temperature = props.float_("temperature");
+        if (props.has_property("scale"))
+            m_scale = props.float_("scale");
+        else
+            m_scale = 1.f;
```
Speierers, line 67: You should use `props.float_("scale", 1.f)` here instead of the `if .. else ..` statement

src/spectra/blackbody.cpp
```diff
@@ -108,7 +120,7 @@ class BlackBodySpectrum final : public Texture<Float, Spectrum> {
                        lambda5 = sqr(lambda2) * lambda;
 
             // Wien's approximation to Planck's law
-            Wavelength pdf = 1e-9f * c0 * exp(-c1 / (lambda * m_temperature))
+            Wavelength pdf = 1e-9f * m_scale * c0 * exp(-c1 / (lambda * m_temperature))
```
Speierers, line 123: Do we really need to scale the PDF here? I think those values should be "normalized" no? Is this how it is implemented in Mitsuba 0.6?

Speierers (changes requested):
Hi @alhirzel ,

Thanks for this PR. Please take a look at the two comments below.

Regarding testing, it would be great to add a chi2 test to check that this spectrum remains "valid" after scaling.

Also, should we make sure `m_scale` is always a positive value? Always greater than zero? Maybe you could throw an error in the constructor is this is not the case.

Actually, you might need to check for this in `parameters_changed()` as well in case someone sets a negative value using the `traverse` mechanism.

PR #596, opened 2022-05-12
Krisrst/advanced lenses

krisrst:
**This PR is a work-in-progress.** As of right now this is _not_ ready for "detailed reviewing" - although general feedback is very much appreciated. There are unimplemented methods, logging of rays to csv files, some very hacky solutions, and files that should not go into mitsuba2 master, that I have used "just to prove that this can be done" - this will be fixed as I go. However, at the current stage the development suffices for a basic demonstration.

The goal of this PR is to realize more complex lens simulations that go beyond the basic "thinlens" and "pinhole" camera models. This can be used to evaluate or even optimize lens designs, for example in terms of distortion. This PR introduces three new components to Mitsuba2:

1. A new type of **sensor**, called "cmos". The cmos sensor will have rays **originate at the surface of the sensor array** (focus_p) rather than **on the aperture of the lens** (aperture_p). It is designed to be able to shoot rays directly into a system of lens surfaces. If there is no optical system the cmos sensor will just produce a "white" image (light no longer gets focused at the die).
2. A new shape **asph_surf**. This is an aspheric surface[1] model. It can model hyperbolic, parabolic, spherical and elliptic lenses. The paper also extends the model with a power series which is currently not implemented (but the plan is that it will). The aspheric surface is not a volume; but just a surface that rays can pass through. 
3. A new shape **cylhollow**. It is a hollow cylinder and quite simple compared to the aspheric lens surface. It is used to encapsulate the lens housing.

[1] Schuhmann, Rainer. "Description of aspheric surfaces." Advanced Optical Technologies 8.3-4 (2019): 267-278.

These components can be used to setup a simple biconvex lens:

image

The following figure shows a subsampled collection of points where the shapes detect origin rays and intersections through the lens housing:

image

We can experiment with the lens parameters to verify that it behaves "physically". For example, in this figure, we regulate the **radius of the circle that the "startlens" shoots rays into**. This effectively varies the aperture of the sensor, which is normally done with some physical obstruction between the lens and the sensor (but we can fake that like this):

image

We see that the lens focuses well with an aperture between 0.5 and 2.5.

In the following figure, we vary the **focal length** of the system; i.e. the distance between the lens and the sensor:

image

Where we can see that "optimal" focus is at approximately 47 mm from the first lens. Outside of this region, the image blurs because the rays no longer hit the focal point of the lens system.

A single high resolution image of the rendered scene:
image

**Compiling / Testing **

Mitsuba.conf
```
    "enabled": [
        # The "scalar_rgb" variant *must* be included at the moment.
        "scalar_rgb",
        "scalar_spectral",
        "packet_rgb",
        "packet_spectral",
        "gpu_rgb",
        "gpu_spectral"
    ],
```
The "main test script" is located in a temporary folder "temp/test.xml". It reuses parts from the `cbox-rgb.xml` script. To run the demo:

`/build/dist/mitsuba -D flen=47.0 -D fov=4 -D aperture=12.5 -D res=256 -D spp=128 -D max_depth=12 -m scalar_rgb -o test temp/test.xml -t 16 2> /dev/null`

At the time of writing I use the `stderr` file descriptor to subsample ray points etc. Any suggestions on alternative ways to "debug rays" is very welcome.

**ToDo List**

- GPU support
- Color filter array for CMOS sensor?
- Implement missing methods
- Fix bounding box calculation
- Fix surface area calculation
- Generic cleanup code
- Extend aspheric surface with power series

*Please add the labels (e.g. bug, feature, ..) corresponding to this PR*

## Description

*Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.*

Fixes # (issue)

## Testing

*Please describe the tests that you added to verify your changes.*

## Checklist:

*Please make sure to complete this checklist before requesting a review.*

- [ ] My code follows the [style guidelines](https://mitsuba2.readthedocs.io/en/latest/src/developer_guide/intro.html#introduction) of this project
- [ ] My changes generate no new warnings
- [ ] My code also compiles for `gpu_*` and `packet_*` variants. If you can't test this, please leave below
- [ ] I have commented my code
- [ ] I give permission that the Mitsuba 2 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba2/blob/master/LICENSE)


Dexter575:
Hello, this is quite interesting and would love to follow up on this topic.

